PR #6, opened 2016-05-02
Bug fixes and minor refactoring

sbellan:
- Buffer to InputStream not reading values greater that 128
- AccessMask re-structuring
- Expose SID constructor and fields, needed when transforming form one SID struture to another(jcifs)
- Workaround MAX_OUTPUT_BUFFER_LENGTH issues observed in Windows 10
- Make some attributes accessible for use in API's


src/main/java/com/hierynomus/protocol/commons/buffer/Buffer.java
```diff
@@ -680,7 +680,7 @@ public InputStream asInputStream() {
             @Override
             public int read() throws IOException {
                 try {
-                    return Buffer.this.readByte();
+                    return Buffer.this.readByte() & 0xFF;
```
hierynomus: Can you add a test for this?

src/main/java/com/hierynomus/smbj/smb2/messages/SMB2IoctlRequest.java
```diff
@@ -59,15 +57,24 @@ protected void writeTo(SMBBuffer smbBuffer) {
...
 
         int offset = SMB2Header.STRUCTURE_SIZE + 56;
-        smbBuffer.putUInt32(offset); // InputOffset (4 bytes)
-        smbBuffer.putUInt32(inputData.length); // InputCount (4 bytes)
-        smbBuffer.putUInt32(MAX_OUTPUT_BUFFER_LENGTH); // MaxInputResponse (4 bytes)
+        if (inputData.length > 0) {
+            smbBuffer.putUInt32(offset); // InputOffset (4 bytes)
+            smbBuffer.putUInt32(inputData.length); // InputCount (4 bytes)
+        } else {
+            smbBuffer.putUInt32(0); // InputOffset (4 bytes)
```
hierynomus: Must the offset be '0'? Or should it point to the location where the input would have been?

src/main/java/com/hierynomus/smbj/smb2/messages/SMB2IoctlRequest.java
```diff
@@ -59,15 +57,24 @@ protected void writeTo(SMBBuffer smbBuffer) {
...
-
+        // TODO 3.3.5.15 Receiving an SMB2 IOCTL Request
+        // Spent days on the INVALID_PARAMETER error which turns out
+        // due to the validation specified in the above section, but not sure which one.
+        // If dialect is 3.x then setting to MAX_OUTPUT_BUFFER_LENGTH works else it fails with INVALID_PARAMETER
+        // To make it work in 2.0.2, subtract the (inputdata length + 8).
+        smbBuffer.putUInt32(MAX_OUTPUT_BUFFER_LENGTH - (inputData.length + 8)); // MaxOutputResponse (4 bytes)
         smbBuffer.putUInt32(fsctl ? 1 : 0); // Flags (4 bytes)
-        smbBuffer.putReserved4(); // Reserved2 (4 bytes)
+        smbBuffer.putReserved4(); // Reserved (4 bytes)
```
hierynomus: The field is actually called `Reserved2` in the spec (because it is the second reserved field of the message.

hierynomus:
Few comments :)

src/main/java/com/hierynomus/smbj/smb2/messages/SMB2IoctlRequest.java
hierynomus: Ok, just read up on this field. The question is, how to interpret this:

```
The offset, in bytes, from the beginning of the SMB2 header to the input data buffer. If no input data is required for the FSCTL/IOCTL command being issued, the client SHOULD set this value to 0.
```

Do we know when no input data is required, do we need to add some check to verify the conditions when inputdata is required?

src/main/java/com/hierynomus/protocol/commons/buffer/Buffer.java
sbellan: The second test in  src/test/groovy/com/hierynomus/spnego/NegTokenInitTest.groovy covers this.

hierynomus: I would prefer a unit test case for the BufferTest. Now it is very obscure :)

src/main/java/com/hierynomus/smbj/smb2/messages/SMB2IoctlRequest.java
sbellan: The Spec says,

The following FSCTL requests do not provide an input buffer: 􏰀 FSCTL_PIPE_PEEK
􏰀 FSCTL_PIPE_WAIT
􏰀 FSCTL_PIPE_TRANSCEIVE
􏰀 FSCTL_SRV_ENUMERATE_SNAPSHOTS
􏰀 FSCTL_SRV_REQUEST_RESUME_KEY
􏰀 FSCTL_QUERY_NETWORK_INTERFACE_INFO

Not sure if we need to check that inputData is empty for these request and non empty for all other requests. If you strongly feel it should be there, I can add it.

src/main/java/com/hierynomus/protocol/commons/buffer/Buffer.java
sbellan: Yes, it is very obscure, it took me some time to get my head around it. When a cast happens from byte to int, if the high bit of the byte is 1 then the int seems to have the upper byte as all 1's. I've added a BufferTest for it.

Btw, Just looked at ByteArrayInputStream source,
    public synchronized int read() {
        return (pos < count) ? (buf[pos++] & 0xff) : -1;
    }

PR #9, opened 2016-05-14
Expose more attributes

sbellan:
Expose some private attributes and don't use some of the new iterators of bouncycastle

### DIFF
--- a/src/main/java/com/hierynomus/msdtyp/SID.java
+++ b/src/main/java/com/hierynomus/msdtyp/SID.java
@@ -16,6 +16,7 @@
 package com.hierynomus.msdtyp;
 
 import com.hierynomus.protocol.commons.ByteArrayUtils;
+import com.hierynomus.protocol.commons.EnumWithValue;
 import com.hierynomus.protocol.commons.buffer.Buffer;
 import com.hierynomus.smbj.common.SMBBuffer;
 
@@ -24,19 +25,35 @@ import com.hierynomus.smbj.common.SMBBuffer;
  */
 public class SID {
 
-    public static final int SID_TYPE_NONE = 0;
-    public static final int SID_TYPE_USER = 1;
-    public static final int SID_TYPE_DOM_GRP = 2;
-    public static final int SID_TYPE_DOMAIN = 3;
-    public static final int SID_TYPE_ALIAS = 4;
-    public static final int SID_TYPE_WKN_GRP = 5;
-    public static final int SID_TYPE_DELETED = 6;
-    public static final int SID_TYPE_INVALID = 7;
-    public static final int SID_TYPE_UNKNOWN = 8;
-    public static final int SID_TYPE_COMPUTER = 8;
-    public static final int SID_TYPE_LABEL = 10;
+    public enum SidType implements EnumWithValue<SidType> {
+        SID_TYPE_NONE(0, "0"),
+        SID_TYPE_USER(1, "User"),
+        SID_TYPE_DOM_GRP(2, "Domain group"),
+        SID_TYPE_DOMAIN(3, "Domain"),
+        SID_TYPE_ALIAS(4, "Local group"),
+        SID_TYPE_WKN_GRP(5, "Builtin group"),
+        SID_TYPE_DELETED(6, "Deleted"),
+        SID_TYPE_INVALID(7, "Invalid"),
+        SID_TYPE_UNKNOWN(8, "Unknown"),
+        SID_TYPE_COMPUTER(9, "Computer"),
+        SID_TYPE_LABEL(10, "Label");
 
-    static final String[] SID_TYPE_NAMES = new String[]{"0", "User", "Domain group", "Domain", "Local group", "Builtin group", "Deleted", "Invalid", "Unknown", "Computer", "Label"};
+        private long value;
+        private String name;
+
+        SidType(long value, String name) {
+            this.value = value;
+            this.name = name;
+        }
+
+        public long getValue() {
+            return value;
+        }
+
+        public String getName() {
+            return name;
+        }
+    }
 
     public static SID EVERYONE = new SID((byte) 1, new byte[]{0, 0, 0, 0, 0, 1}, new long[]{0});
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/hierynomus/msdtyp/SID.java
```diff
@@ -23,6 +23,21 @@
...
+    public static final int SID_TYPE_DOMAIN = 3;
+    public static final int SID_TYPE_ALIAS = 4;
+    public static final int SID_TYPE_WKN_GRP = 5;
+    public static final int SID_TYPE_DELETED = 6;
+    public static final int SID_TYPE_INVALID = 7;
+    public static final int SID_TYPE_UNKNOWN = 8;
+    public static final int SID_TYPE_COMPUTER = 8;
+    public static final int SID_TYPE_LABEL = 10;
+
+    static final String[] SID_TYPE_NAMES = new String[]{"0", "User", "Domain group", "Domain", "Local group", "Builtin group", "Deleted", "Invalid", "Unknown", "Computer", "Label"};
```
hierynomus: Why not an ENUM?

src/main/java/com/hierynomus/spnego/NegTokenInit.java
```diff
@@ -102,11 +103,18 @@ public NegTokenInit read(byte[] bytes) throws IOException {
     public NegTokenInit read(Buffer<?> buffer) throws IOException {
         try {
             ASN1Primitive applicationSpecific = new ASN1InputStream(buffer.asInputStream()).readObject();
-            if (!(applicationSpecific instanceof ASN1ApplicationSpecific)) {
+            if (!(applicationSpecific instanceof BERApplicationSpecific || applicationSpecific instanceof DERApplicationSpecific)) {
```
hierynomus: Why do we need compat with earlier BC?

src/main/java/com/hierynomus/msdtyp/SID.java
sbellan: Yes, this could be an enum.

src/main/java/com/hierynomus/spnego/NegTokenInit.java
sbellan: Related to our project where we are unable to move to the new BC because of another dependency. Since there is no change to the functionality, thought it might be ok.

hierynomus: Aren't there CVE's against 1.51?

sbellan: Smbj will continue to use 1.54 and advertise the dependency as 1.54. Just that when someone is using smbj along with some other dependency and understands the risk and wants to forcefully change the boucycastle dependency to version 1.51 in their own builds. If we cant get this, then we wont be able to use the master version for some time.

hierynomus: ok, agreed. I actually want to see whether for the ASN.1 part we can't use smetihing else, as it is the only hardcoded dependency on BC. But that is for a later date :)

src/main/java/com/hierynomus/msdtyp/SID.java
```diff
@@ -16,13 +16,45 @@
...
 
 /**
  * [MS-DTYP].pdf 2.4.2 SecurityIdentifier SID
  */
 public class SID {
+
+    public enum SidType implements EnumWithValue<SidType> {
+        SID_TYPE_NONE(0, "0"),
+        SID_TYPE_USER(1, "User"),
+        SID_TYPE_DOM_GRP(2, "Domain group"),
```
hierynomus: Rename this to `SID_TYPE_DOMAIN_GROUP` (Unless the current name is the actual name used in the documentation)

src/main/java/com/hierynomus/msdtyp/SID.java
```diff
@@ -16,13 +16,45 @@
...
+        SID_TYPE_DOMAIN(3, "Domain"),
+        SID_TYPE_ALIAS(4, "Local group"),
+        SID_TYPE_WKN_GRP(5, "Builtin group"),
```
hierynomus: Same here, is this the name in the documentation? Else rename it.

sbellan: Can you just merge this change. I've made more changes after this and I will make this change after those. Thanks.

hierynomus:
ok, will merge. Comments are there to point out the coding guideline. Keeping names close to the documentation alows for easy lookup and bugfixes ;)

src/main/java/com/hierynomus/spnego/NegTokenInit.java
krivanekm-javlin: Is there any chance of replacing ASN.1 from BouncyCastle with something else? BouncyCastle dependency is a bit problematic because of US strong cryptography export restrictions: https://en.wikipedia.org/wiki/Strong_cryptography#Legal_issues

We have solved this by making our users download BC separately and add it to our application classpath, but it complicates the installation a bit.

What's the current status of https://github.com/hierynomus/asn-one?

hierynomus: Indeed, asn-one is underway. Need to put some focus time on it to finish it. Can't even remember what was holding it back 😉 ...
Feel free to submit a PR to migrate to asn-one based on the current master build. Then we can see quickly enough what's still missing!

hierynomus: @krivanekm-javlin Just created PR #394. It is working for the integration tests. Just need to release a final version of asn-one now.

krivanekm-javlin: Wow, that was fast! I started working on it and I think I have the changes ready, but you were faster. :)

What does this mean? Will BC still be a required dependency? There are references in com.hierynomus.security.bc package. Are those required or optional at runtime?

In the case it's still required:
Since Java 8u161, unlimited cryptography is enabled by default:
https://www.oracle.com/technetwork/java/javase/8u161-relnotes-4021379.html#JDK-8170157 Would this allow making BouncyCastle optional when running with unlimited cryptography?

Thanks for your reply.

hierynomus: @krivanekm-javlin Let's continue the discussion on PR #394.

PR #10, opened 2016-05-26
High level API

sbellan:
Another go at the high level API. Please review and feel free to rename,refactor.


hierynomus:
Thanks!
I'll merge this as it is a long way in the right direction. But we definitely need to do some more refactoring ;)

PR #26, opened 2016-11-07
Basic SMBv2 signing implementation

ckherrmann:
Implement SMBv2 message signing. We extract the signing key from the NTLM authentication, perform proper keyexchange with the server, and sign most messages sent to the server. We also check signature of incoming messages, if the Connection requires it.

Make Connection output Buffered, so that when we write to the connection, a whole packet is written all at once.  This makes it much easier to use a protocol analyzer.  Previously, a lot of the time a packet would be broken up.

I've tested this very lightly against a Windows 2012 server.

### DIFF
--- a/src/main/java/com/hierynomus/mssmb2/SMB2Header.java
+++ b/src/main/java/com/hierynomus/mssmb2/SMB2Header.java
@@ -28,6 +28,8 @@ import static com.hierynomus.smbj.connection.NegotiatedProtocol.SINGLE_CREDIT_PA
  */
 public class SMB2Header {
     public static final int STRUCTURE_SIZE = 64;
+    public static final int SIGNATURE_OFFSET = 48;
+    public static final int SIGNATURE_SIZE = 16;
 
     private SMB2Dialect dialect;
     private int creditCharge = 1;
@@ -137,6 +139,10 @@ public class SMB2Header {
         this.dialect = dialect;
     }
 
+    public boolean isFlagSet(SMB2MessageFlag flag) {
+        return ((this.flags & flag.getValue()) != 0);
+    }
+    
     public void setFlag(SMB2MessageFlag flag) {
         this.flags |= flag.getValue();
     }
@@ -204,5 +210,24 @@ public class SMB2Header {
 
     public void setCreditCharge(int creditCharge) {
         this.creditCharge = creditCharge;
+    }
+    
+    public String toString() {
+        return String.format(
+        "dialect=%s, creditCharge=%s, creditRequest=%s, creditResponse=%s, message=%s, messageId=%s, asyncId=%s, sessionId=%s, treeId=%s, status=%s, statusCode=%s, flags=%s, nextCommandOffset=%s",
+        dialect,
+        creditCharge,
+        creditRequest,
+        creditResponse,
+        message,
+        messageId,
+        asyncId,
+        sessionId,
+        treeId,
+        status,
+        statusCode,
+        flags,
+        nextCommandOffset);
+
     }
 }
--- a/src/main/java/com/hierynomus/mssmb2/SMB2Header.java
+++ b/src/main/java/com/hierynomus/mssmb2/SMB2Header.java
@@ -141,7 +141,7 @@ public class SMB2Header {
     }
 
     public boolean isFlagSet(SMB2MessageFlag flag) {
-        return EnumWithValue.EnumUtils.isSet(this.flags, flag);
+        return isSet(this.flags, flag);
     }
     
     public void setFlag(SMB2MessageFlag flag) {
--- a/src/main/java/com/hierynomus/mssmb2/SMB2Header.java
+++ b/src/main/java/com/hierynomus/mssmb2/SMB2Header.java
@@ -16,6 +16,7 @@
 package com.hierynomus.mssmb2;
 
 import com.hierynomus.mserref.NtStatus;
+import com.hierynomus.protocol.commons.EnumWithValue;
 import com.hierynomus.protocol.commons.buffer.Buffer;
 import com.hierynomus.smbj.common.SMBBuffer;
 
@@ -140,7 +141,7 @@ public class SMB2Header {
     }
 
     public boolean isFlagSet(SMB2MessageFlag flag) {
-        return ((this.flags & flag.getValue()) != 0);
+        return EnumWithValue.EnumUtils.isSet(this.flags, flag);
     }
     
     public void setFlag(SMB2MessageFlag flag) {
--- a/src/main/java/com/hierynomus/mssmb2/messages/SMB2SessionSetup.java
+++ b/src/main/java/com/hierynomus/mssmb2/messages/SMB2SessionSetup.java
@@ -59,9 +59,10 @@ public class SMB2SessionSetup extends SMB2Packet {
         buffer.putUInt32(clientCapabilities & 0x01); // Capabilities (4 bytes) (only last byte can be set)
         buffer.putReserved4(); // Channel (4 bytes)
         buffer.putUInt16(SMB2Header.STRUCTURE_SIZE + 25 - 1); // SecurityBufferOffset (2 bytes) (header structure size + Session setup structure size - 1)
-        buffer.putUInt16(securityBuffer.length); // SecurityBufferLength (2 bytes)
+        buffer.putUInt16((securityBuffer!=null)?securityBuffer.length:0); // SecurityBufferLength (2 bytes)
         buffer.putUInt64(previousSessionId); // PreviousSessionId (8 bytes)
-        buffer.putRawBytes(securityBuffer); // SecurityBuffer (variable)
+        if (securityBuffer != null)
+            buffer.putRawBytes(securityBuffer); // SecurityBuffer (variable)
     }
 
     @Override
@@ -82,7 +83,9 @@ public class SMB2SessionSetup extends SMB2Packet {
             buffer.rpos(securityBufferOffset);
             return buffer.readRawBytes(securityBufferLength);
         }
-        throw new SMBRuntimeException("The SMB2 Session Setup response should contain a positive length security buffer");
+        else
+            return new byte[0];
+//        throw new SMBRuntimeException("The SMB2 Session Setup response should contain a positive length security buffer");
     }
 
     private void putFlags(SMBBuffer buffer) {
--- a/src/main/java/com/hierynomus/mssmb2/messages/SMB2SessionSetup.java
+++ b/src/main/java/com/hierynomus/mssmb2/messages/SMB2SessionSetup.java
@@ -59,10 +59,11 @@ public class SMB2SessionSetup extends SMB2Packet {
         buffer.putUInt32(clientCapabilities & 0x01); // Capabilities (4 bytes) (only last byte can be set)
         buffer.putReserved4(); // Channel (4 bytes)
         buffer.putUInt16(SMB2Header.STRUCTURE_SIZE + 25 - 1); // SecurityBufferOffset (2 bytes) (header structure size + Session setup structure size - 1)
-        buffer.putUInt16((securityBuffer!=null)?securityBuffer.length:0); // SecurityBufferLength (2 bytes)
+        buffer.putUInt16((securityBuffer!=null) ? securityBuffer.length : 0); // SecurityBufferLength (2 bytes)
         buffer.putUInt64(previousSessionId); // PreviousSessionId (8 bytes)
-        if (securityBuffer != null)
+        if (securityBuffer != null) {
             buffer.putRawBytes(securityBuffer); // SecurityBuffer (variable)
+        }
     }
 
     @Override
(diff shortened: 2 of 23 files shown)


src/main/java/com/hierynomus/mssmb2/SMB2Header.java
```diff
@@ -137,6 +139,10 @@ public void setDialect(SMB2Dialect dialect) {
         this.dialect = dialect;
     }
 
+    public boolean isFlagSet(SMB2MessageFlag flag) {
+        return ((this.flags & flag.getValue()) != 0);
```
hierynomus: `EnumWithValue.EnumUtils.isSet(this.flags, flag);`

src/main/java/com/hierynomus/mssmb2/messages/SMB2SessionSetup.java
```diff
@@ -59,9 +59,10 @@ protected void writeTo(SMBBuffer buffer) {
         buffer.putUInt32(clientCapabilities & 0x01); // Capabilities (4 bytes) (only last byte can be set)
         buffer.putReserved4(); // Channel (4 bytes)
         buffer.putUInt16(SMB2Header.STRUCTURE_SIZE + 25 - 1); // SecurityBufferOffset (2 bytes) (header structure size + Session setup structure size - 1)
-        buffer.putUInt16(securityBuffer.length); // SecurityBufferLength (2 bytes)
+        buffer.putUInt16((securityBuffer!=null)?securityBuffer.length:0); // SecurityBufferLength (2 bytes)
```
hierynomus: Spacing around operators

src/main/java/com/hierynomus/mssmb2/messages/SMB2SessionSetup.java
```diff
@@ -59,9 +59,10 @@ protected void writeTo(SMBBuffer buffer) {
...
         buffer.putUInt64(previousSessionId); // PreviousSessionId (8 bytes)
-        buffer.putRawBytes(securityBuffer); // SecurityBuffer (variable)
+        if (securityBuffer != null)
```
hierynomus: Please always use curly braces, even for 1-line statements

src/main/java/com/hierynomus/mssmb2/messages/SMB2SessionSetup.java
```diff
@@ -82,7 +83,9 @@ protected void readMessage(SMBBuffer buffer) throws Buffer.BufferException {
             buffer.rpos(securityBufferOffset);
             return buffer.readRawBytes(securityBufferLength);
         }
-        throw new SMBRuntimeException("The SMB2 Session Setup response should contain a positive length security buffer");
+        else
```
hierynomus: See above.

src/main/java/com/hierynomus/smbj/auth/NtlmAuthenticator.java
```diff
@@ -60,19 +61,24 @@ public NtlmAuthenticator create() {
         }
     }
 
-    public long authenticate(Connection connection, AuthenticationContext context) throws TransportException {
+    public Session authenticate(Connection connection, AuthenticationContext context) throws TransportException {
```
hierynomus: It would be nicer to introduce a `SessionDetails` object which encapsulates the SigningKey and the SessionId. And leave the creation of the session to the connection. Now you creating a(nother) package cycle (I'm trying to break them at this moment).

src/main/java/com/hierynomus/smbj/connection/Connection.java
```diff
@@ -152,6 +175,9 @@ public Session authenticate(AuthenticationContext authContext) {
 
             Request request = new Request(packet.getHeader().getMessageId(), UUID.randomUUID(), packet);
             connectionInfo.getOutstandingRequests().registerOutstanding(request);
+            if (signingKey != null)
```
hierynomus: curlies :)

src/main/java/com/hierynomus/smbj/connection/Connection.java
```diff
@@ -217,6 +243,40 @@ public void handle(SMB2Packet packet) throws TransportException {
...
+            
+            // check packet signature.  Drop the packet if it is not correct.
+            if (session.getSigningRequired()) {
+                if (packet.getHeader().isFlagSet(SMB2MessageFlag.SMB2_FLAGS_SIGNED)) {
+                    packet.getBuffer().rpos(0);
+                    if (!MessageSigning.validateSignature(packet.getBuffer().array(), packet.getBuffer().available(), session.getSigningKey())) {
+                        logger.warn("Invalid packet signature");
+//                        return;
+                    }
+                }
```
hierynomus: See the code style throughout the code, else on same line as `}`

src/main/java/com/hierynomus/smbj/connection/Connection.java
```diff
@@ -217,6 +243,40 @@ public void handle(SMB2Packet packet) throws TransportException {
...
+                else {
+                    logger.warn("Illegal request, session requires message signing, but the message is not signed.");
```
hierynomus: Might be better to make this tuneable to just a warning as it is now, or a hard fail.

src/main/java/com/hierynomus/smbj/connection/Connection.java
```diff
@@ -217,6 +243,40 @@ public void handle(SMB2Packet packet) throws TransportException {
...
+                    return;
+                }
+            }
+            
+            // check packet signature.  Drop the packet if it is not correct.
+            if (session.getSigningRequired()) {
+                if (packet.getHeader().isFlagSet(SMB2MessageFlag.SMB2_FLAGS_SIGNED)) {
+                    packet.getBuffer().rpos(0);
+                    if (!MessageSigning.validateSignature(packet.getBuffer().array(), packet.getBuffer().available(), session.getSigningKey())) {
+                        logger.warn("Invalid packet signature");
```
hierynomus: Might be better to make this tuneable to just a warning as it is now, or a hard fail.

src/main/java/com/hierynomus/smbj/share/DiskShare.java
```diff
@@ -318,7 +318,7 @@ private void deleteCommon(String path, SMB2CreateRequest smb2CreateRequest)
         Connection connection = session.getConnection();
 
         // TODO Use Compounding
-        Future<SMB2CreateResponse> sendFuture = connection.send(smb2CreateRequest);
+        Future<SMB2CreateResponse> sendFuture = connection.send(smb2CreateRequest, session.getSigningRequired()?session.getSigningKey():null);
```
hierynomus: Create a `Session.send(...)` method that takes just the packet to send. And delegate that to the `connection.send(packet, key)` call.

src/main/java/com/hierynomus/smbj/transport/BaseTransport.java
```diff
@@ -57,5 +61,39 @@ public void write(SMB2Packet packet) throws TransportException {
...
+        } finally {
+            writeLock.unlock();
+        }
+    }
+
     protected abstract void doWrite(SMBBuffer packetData) throws IOException;
+
+    
+    // [MS-SMB2] 3.1.4.1 Signing An Outgoing Message
+    // If Connection.Dialect is "2.0.2" or "2.1", the sender MUST compute a 32-byte hash using HMAC-SHA256 over the entire message,
```
hierynomus: Maybe immediately make the provision for signing with a higher dialect?

src/main/java/com/hierynomus/smbj/common/MessageSigning.java
```diff
@@ -0,0 +1,108 @@
...
+     *            the session's signing key.
+     * @throws InvalidKeyException
+     * @throws NoSuchAlgorithmException
+     */
+    private static byte[] computeSignature(byte[] buffer, int len, byte[] signingKey) throws NoSuchAlgorithmException,
+                    InvalidKeyException {
+        if (len < SMB2Header.STRUCTURE_SIZE)
+            throw new IllegalArgumentException("Buffer must be longer than 64 bytes");
+        SecretKeySpec signingKeySpec = new SecretKeySpec(signingKey, MessageSigning.HMAC_SHA256_ALGORITHM);
+        Mac mac = Mac.getInstance(MessageSigning.HMAC_SHA256_ALGORITHM);
```
hierynomus: If we pass the signingKey in here as a SecretKeySpec, I think you can do `Mac.getInstance(signingKey.getAlgorithm())`?

src/main/java/com/hierynomus/smbj/connection/Connection.java
ckherrmann: Do you mean the tunable parameter should be a field in the Config class?  And is it your intention to have a bunch of these kinds of items?  If so, let me know what kind of architectural style you'd like to use.  For instance, it could be a boolean method (like the other Config items) or it could be a System.getProperty() kind of thing.

hierynomus:
A parameter in the config object would be best. I don't like
System.getProperty settings, as they cannot be tuned per connection. I
could very well imagine the library being used in a single jvm to connect
to multiple different hosts using different configurations. (it's how we
use it at xebialabs ;))

Op 7 nov. 2016 11:55 p.m. schreef "ckherrmann" <notifications@github.com>:

src/main/java/com/hierynomus/smbj/transport/BaseTransport.java
ckherrmann: This is my first foray into your codebase, so I didn't want to make assumptions about how you'd like to design that.

hierynomus:
As for the config parameter itself, maybe something like `isStrictSigning`?
Cant come up with a better name at this moment.

Op 7 nov. 2016 11:59 p.m. schreef "Jeroen van Erp" <jeroen@javadude.nl>:

src/main/java/com/hierynomus/smbj/common/MessageSigning.java
ckherrmann: What benefit are you thinking about?  I haven't looked at what is required to do SMBv3 signing, so I can't say if this would be sufficient to avoid having to rewrite the signing code.  I kind of assumed I would have to write a MessageSigningv3 class or something.

hierynomus:
Got your point. The problem now is, is that nothing keels over when we
start adding smb3x stuff. Would be good at least to get an exception then
that signing has not been implemented for that dialect.

Op 8 nov. 2016 12:00 a.m. schreef "ckherrmann" <notifications@github.com>:

hierynomus:
Neitjer have I looked at it in depth. I do remember the algorithm changes.
Also it is safe to pass around the secretkeyspec instead of the bare
byte[]. Makes it a bit cleaner and more 'typed'.

Op 8 nov. 2016 12:04 a.m. schreef "ckherrmann" <notifications@github.com>:

src/main/java/com/hierynomus/smbj/auth/NtlmAuthenticator.java
ckherrmann: Here's an issue.  In the NtlmAuthenticator, this can make sense because there is only a simple exchange.  However, in a more general GSSAPI authentication scenario where you have multiple rounds, it seems like the suggestion from the MS-SMB2 document is that we should have a separate session table for sessions that are in "authentication state", but that they are full session objects.  I don't fully understand why that is, but I haven't looked carefully at the situation.

So, while I understand the idea of separating SessionDetails from Session, I'm not convinced that this is the best separation.  So I'd like to punt this decision until we understand this better.

My gut feeling is that this is all driven from a (possibly unavoidable) messy protocol design, where we have a "stack" of protocol contexts (eg, connection, session, treeconnect, share, etc.) in which all actions will take place.  So that means that all of those contexts will need to be split into "details" and "code/state" objects, if we really care about the architectural separation.  Otherwise, we just accept the cycle, on the basis that all elements of the stack are actually required by the protocol (and by Java limitations).

hierynomus: Still I think the Authenticator is the wrong place to create the session. The connection builds the session, the authentication takes care of authenticating the session. One other option is that we pass in an `UnauthenticatedSession` object that constructs the 'real' `Session`. Instead of the `AuthenticationContext` that we now pass in.

The fact that you had to 'open up' classes that were made package private is to me always an immediate smell.

hierynomus:
Hi @ckherrmann. I'll merge this PR as is, and will do the last updates as I have some time today. Thanks for the great contribution. Looking forward to receiving more when you have time!

src/main/java/com/hierynomus/smbj/auth/NtlmAuthenticator.java
ckherrmann: I was trying not to rearchitect too much from your original design.  That said, I don't disagree with your comment.  But does that block the pull request, or can we deal with that later?

ckherrmann:
OK, thanks  I have a SPNEGO implementation on a branch, but I kind of want to discuss the best way to show it using a test app before doing a PR.

hierynomus:
Let move that discussion to #28 :)

PR #30, opened 2016-11-12
SPNEGO implementation for Kerberos authentication using Kerberos Subject and GSSCredential

ckherrmann:
To use, obtain a Subject and extract its GSSCredential, and use these to create a GSSAuthenticationContext that you can give to the Connection.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,33 @@
+image: java:8-jdk
+
+stages:
+  - build
+  - test
+  - deploy
+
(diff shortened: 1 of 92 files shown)


hierynomus:
Thanks. Will have a look on Monday :)

hierynomus:
I'll merge this. One thing we need to do is look into how we can test this. There are some GSS tests in https://github.com/hierynomus/sshj. Maybe you can have a look at this to see whether we can write some tests for it here also.

PR #32, opened 2016-11-29
Misc Fixes

sbellan:
- SID equals/hashCode
- More NtStatus defined
- Add the actual NtStatus in the api exception
- Directory list to consume the full output


src/it/java/SmbjTest.java
```diff
@@ -333,6 +333,15 @@ public void testRpc() throws IOException, SMBApiException, URISyntaxException {
     }
 
 
+    @Test
+    public void testSID() {
```
hierynomus: Could you move this test to the `SIDTest.groovy` class? Also please don't println, just assert :)

PR #38, opened 2017-03-15
Pull request to fix NegTokenInit.parseTagged method

niggemannmj:
This should resolve issue 37 which I have observed with Samba shares on RHEL 6.


hierynomus:
Thanks! If you have a sample byte string which contains these fields, we could also add a test in the `NegTokenInitTest`. That would complete the PR :)

PR #49, opened 2017-06-02
Direct seekable buffer read/write support

sopak:
Sometimes it is useful to access files randomly.
I have implemented java.nio interface for Smbj.
It is buffer oriented.

### DIFF
--- a/src/main/java/com/hierynomus/smbj/share/File.java
+++ b/src/main/java/com/hierynomus/smbj/share/File.java
@@ -102,7 +102,6 @@ public class File extends DiskEntry {
         return new FileOutputStream(this, listener);
     }
 
-
     /**
      * Direct buffer read from the file
      *
@@ -143,14 +142,12 @@ public class File extends DiskEntry {
                 throw new SMBApiException(smb2Packet.getHeader(), "Read failed for " + this);
             }
             int read = smb2Packet.getDataLength();
-            if(read>payloadSize){
-                read=payloadSize;
+            if (read > payloadSize) {
+                read = payloadSize;
             }
             System.arraycopy(smb2Packet.getData(), 0, dst, 0, read);
             return read;
-        } catch (InterruptedException e) {
-            throw new IOException(e);
-        } catch (ExecutionException e) {
+        } catch (InterruptedException | ExecutionException e) {
             throw new IOException(e);
         }
     }
@@ -160,11 +157,11 @@ public class File extends DiskEntry {
      *
      * @param src        source buffer to be  written
      * @param fileOffset offset where to start write
-     * @param lenght     length of src buffer
+     * @param length     length of src buffer
      * @return bytes written
      * @throws IOException
      */
-    public int write(final byte[] src, final int fileOffset, final int lenght) throws IOException {
+    public int write(final byte[] src, final int fileOffset, final int length) throws IOException {
 
         if ((this.accessMask & (AccessMask.GENERIC_WRITE.getValue() | AccessMask.FILE_WRITE_EA.getValue())) == 0) {
             throw new SMBApiException(
@@ -178,37 +175,33 @@ public class File extends DiskEntry {
         Session session = treeConnect.getSession();
         NegotiatedProtocol negotiatedProtocol = session.getConnection().getNegotiatedProtocol();
 
-        ByteChunkProvider provider = new ByteChunkProvider()
-        {
-            int remaining = lenght;
+        ByteChunkProvider provider = new ByteChunkProvider() {
+            int remaining = length;
 
-            public ByteChunkProvider withOffset(int offset){
+            ByteChunkProvider withOffset(int offset) {
                 this.offset = offset;
                 return this;
             }
 
             @Override
-            public boolean isAvailable()
-            {
+            public boolean isAvailable() {
                 return remaining > 0;
             }
 
             @Override
-            protected int getChunk(byte[] chunk) throws IOException
-            {
+            protected int getChunk(byte[] chunk) throws IOException {
                 int write = chunk.length;
                 if (write > remaining) {
                     write = remaining;
                 }
-                System.arraycopy(src, lenght - remaining, chunk, 0, write);
+                System.arraycopy(src, length - remaining, chunk, 0, write);
                 remaining -= write;
 
                 return write;
             }
 
             @Override
-            public int bytesLeft()
-            {
+            public int bytesLeft() {
                 return remaining;
             }
         }.withOffset(fileOffset);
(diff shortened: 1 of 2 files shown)


yin19941005:
Hi @sopak ,

Nice work! I believe this helps a lot! However, I find a small bug on it. [On this line](https://github.com/hierynomus/smbj/pull/49/files#diff-37771c25f310d7ce8f390c4b585748c2R134), the `smb2Packet.getData()` may null and cause the null pointer exception. I tried to open the file with `AccessMask.GENERIC_WRITE` only (no read access) and try to read the file and this line cause the null pointer exception. I further test for the write file without the write access, the exception is already caught.

I suggest to fix this issue by checking the `smb2Packet.getData()` is null or not first and log the NTSTATUS when it is null for debugging. The code as follow :  
```
if(smb2Packet.getData() != null) {
       System.arraycopy(smb2Packet.getData(), 0, dst, 0, read);
}else {
       logger.info("smb2Packet.getData() is null, NTSTATUS = " + smb2Packet.getHeader().getStatus().toString());
}
```

Do you agree with it?

sopak:
@yin19941005 I agree, but I will fix it by checking access mask for write/read permissions.

> 
        if ((this.accessMask & (AccessMask.GENERIC_READ.getValue() | AccessMask.FILE_READ_EA.getValue())) == 0) {
            throw new SMBApiException(
                NtStatus.STATUS_ACCESS_DENIED,
                NtStatus.STATUS_ACCESS_DENIED.getValue(),
                null,
                "The file is not open for reading"
            );
        }

Do you agree with it?

yin19941005:
Hi @sopak ,

I agree we can handle the case that the granted access is not allow for the read/write operation by your new proposed commit. However, I think we should check for read operation is succeed or not. (For the write operation, it's not matter because the server is responsible and it will return error NTSTATUS for us.)

In my opinion, it seems when any error returned on the read request, the `response.getData()` will be null (and may cause the null pointer exception). It make sense because when error occur, the read operation should be aborted. I just read the [MSDN document on here](https://msdn.microsoft.com/en-us/library/cc246729.aspx), there are actually still others possibility to cause error and crash the program by null pointer exception.

I suggest besides your newly commit to prevent the access right error, we should also check the NTSTATUS is success or not before calling the `System.arraycopy(smb2Packet.getData(), 0, dst, 0, read);` : 
```
if (smb2Packet.getHeader() != null) {
  if (smb2Packet.getHeader().getStatus() == NtStatus.STATUS_SUCCESS) {
   System.arraycopy(smb2Packet.getData(), 0, dst, 0, read);
  }
}
```
What's your opinion about this?

sopak:
@yin19941005  smb2Packet.getHeader() is  never null according (I see Erp never checks for it)
> protected final SMB2Header header = new SMB2Header();

what about  this

```
SMB2ReadResponse smb2Packet = send.get();
if (smb2Packet.getHeader().getStatus() != NtStatus.STATUS_SUCCESS) {
    throw new SMBApiException(smb2Packet.getHeader(), "Read failed for " + this);
}
```

src/main/java/com/hierynomus/smbj/share/File.java
```diff
@@ -96,6 +102,143 @@ public OutputStream getOutputStream(final ProgressListener listener) {
...
+        );
+
+        Future<SMB2ReadResponse> send = session.send(rreq);
+        try {
+            SMB2ReadResponse smb2Packet = send.get();
+            if (smb2Packet.getHeader().getStatus() != NtStatus.STATUS_SUCCESS) {
+                throw new SMBApiException(smb2Packet.getHeader(), "Read failed for " + this);
+            }
+            int read = smb2Packet.getDataLength();
+            if(read>payloadSize){
```
hierynomus: formatting is off.

src/main/java/com/hierynomus/smbj/share/File.java
```diff
@@ -96,6 +102,143 @@ public OutputStream getOutputStream(final ProgressListener listener) {
...
+                read=payloadSize;
```
hierynomus: here too

src/main/java/com/hierynomus/smbj/share/File.java
```diff
@@ -96,6 +102,143 @@ public OutputStream getOutputStream(final ProgressListener listener) {
...
+    /**
+     * Direct buffer write to the file
+     *
+     * @param src        source buffer to be  written
+     * @param fileOffset offset where to start write
+     * @param lenght     length of src buffer
+     * @return bytes written
+     * @throws IOException
+     */
+    public int write(final byte[] src, final int fileOffset, final int lenght) throws IOException {
```
hierynomus: Spelling mistake for `length`

src/main/java/com/hierynomus/smbj/share/File.java
```diff
@@ -96,6 +102,143 @@ public OutputStream getOutputStream(final ProgressListener listener) {
...
+            throw new IOException(e);
+        }
+    }
+
+    /**
+     * Direct buffer write to the file
+     *
+     * @param src        source buffer to be  written
+     * @param fileOffset offset where to start write
+     * @param lenght     length of src buffer
```
hierynomus: Spelling mistake for `length`

src/main/java/com/hierynomus/smbj/share/File.java
```diff
@@ -96,6 +102,143 @@ public OutputStream getOutputStream(final ProgressListener listener) {
...
+                NtStatus.STATUS_ACCESS_DENIED.getValue(),
+                null,
+                "The file is not open for writing"
+            );
+        }
+
+        Session session = treeConnect.getSession();
+        NegotiatedProtocol negotiatedProtocol = session.getConnection().getNegotiatedProtocol();
+
+        ByteChunkProvider provider = new ByteChunkProvider()
```
hierynomus: Curlies-open go at end of line. Also for a few more methods.

src/main/java/com/hierynomus/smbj/share/File.java
```diff
@@ -96,6 +102,143 @@ public OutputStream getOutputStream(final ProgressListener listener) {
...
+        {
+            int remaining = lenght;
+
+            public ByteChunkProvider withOffset(int offset){
```
hierynomus: Space before curly

hierynomus:
Thanks for the PR, few minor comments to polish it :)

sopak:
@hierynomus Hi, may you share your formatting settings for this project? 
I will probably add few more implementations.  And I do not want mess it too much ;).

Issue #48, opened 2017-05-23
Support Renames


travisennis:
It appears the library can support renames by using the SMB2SetInfoRequest class, but I'm having trouble figuring out how to configure an instance of that class that will successfully do the rename.

Could you give me some advice on how renames might be implemented? Also, could a rename method be implemented on the Share class, so that the API can hide the complexity of the SMB2SetInfoRequest from application code?

hierynomus:
Yes, the Share should abstract that away. There should be no need to
construct the messages yourself indeed. Will add it.

2017-05-23 20:09 GMT+02:00 Travis Ennis <notifications@github.com>:

travisennis:
Excellent! Thank you! Can you provide an estimate as to when you think it will get added and released?

yin19941005:
Hello,

I had created a pull request for this issue. I tried to move (rename) a "txt" file, empty directory and directory with a "txt" file and all of them passed on my environment. 

@travisennis , could you take a look and test with it?

travisennis:
@yin19941005, I did some testing with your code changes yesterday and everything worked in my environment as well.

hierynomus:
New PR for this is #61

travisennis:
@yin19941005 after further testing, I'm seeing an issue where your solution does not properly rename files or directories that exist inside of other directories.

yin19941005:
Hi @travisennis ,

Could you provide more details information? I am not able to reproduce the case you mentioned. I tested on my environment with the setup below : 

Client : Windows 10 PRO
Server : Windows Server 2008 R2
My test program is a modified version based on [`SmbjTest.java`](https://github.com/hierynomus/smbj/blob/master/src/it/java/SmbjTest.java) : 
```
...... (Skip irrelevant code)

            // Create directories
            share.mkdir(urlFix(writeTest_Path));
            share.mkdir(urlFix(writeTest_Path + "/1"));
            share.mkdir(urlFix(writeTest_Path + "/1/2"));
            share.mkdir(urlFix(writeTest_Path + "/1/2/3"));
            share.mkdir(urlFix(writeTest_Path + "/1/2/3/4"));
            share.mkdir(urlFix(writeTest_Path + "/2"));
            share.mkdir(urlFix(writeTest_Path + "/3"));
            share.mkdir(urlFix(writeTest_Path + "/4"));
            share.mkdir(urlFix(writeTest_Path + "/4/2"));
[...]

PR #52, opened 2017-06-05
Rename file implementation

yin19941005:
- Add the NTstatus for the info class mismatch instead of return unknown.
This should be helpful in debugging and testing.

- Make the securityInformation in SMB2SetInfoRequest using EnumSet and will be consistent with SMB2QueryInfoRequest now.
Just make them consistent.

- Change the Factory on creating the reanmeInfo message and fix the error.
We need "StandardCharsets.UTF_16LE" instead of "StandardCharsets.UTF_16". Using the existing function : ["SMB2Functions.unicode"](https://github.com/hierynomus/smbj/blob/master/src/main/java/com/hierynomus/mssmb2/SMB2Functions.java) to implement it.

- Implement the rename file and rename directory function in DiskShare.
The purpose of this pull request. #48


src/main/java/com/hierynomus/msfscc/fileinformation/FileInformationFactory.java
```diff
@@ -37,7 +38,7 @@
         renBuf.putRawBytes(new byte[]{0, 0, 0, 0, 0, 0, 0});
         renBuf.putUInt64(0);
         renBuf.putUInt32(newName.length() * 2); // unicode
-        renBuf.putRawBytes(newName.getBytes(StandardCharsets.UTF_16));
+        renBuf.putRawBytes(SMB2Functions.unicode(newName));
```
hierynomus: By depending on the `SMB2Functions` you introduce a circular package dependency. Prefer just to use the right `StandardCharsets` constant.

src/main/java/com/hierynomus/smbj/share/DiskShare.java
```diff
@@ -466,6 +465,152 @@ private boolean exists(String path, EnumSet<SMB2CreateOptions> createOptions) th
...
     }
 
+    /***
+     * Rename(Move) Directory. Default will using the false for replaceIfExist option
+     * @param path the path to the original directory
+     * @param newFileName the new directory name/path after rename
+     * @throws TransportException
+     * @throws SMBApiException
+     */
+    public void renameDirectory(String path, String newFileName) throws TransportException, SMBApiException {
```
hierynomus: Here I'm doubting whether we should implement `rename` operations on the `File` and `Directory` objects instead of the `rename(Directory|File)` methods on the DiskShare.

What are your thoughts on this?

src/main/java/com/hierynomus/msfscc/fileinformation/FileInformationFactory.java
yin19941005: Ok, I will change it to `renBuf.putRawBytes(newName.getBytes(StandardCharsets.UTF_16LE)`.

src/main/java/com/hierynomus/smbj/share/DiskShare.java
yin19941005: Yes, maybe you are correct. I tried to use `setRenameFileInfomation(String path, String newFileName, boolean replaceIfExist, EnumSet<SMB2CreateOptions> createOptions)` wtih createOptions null and it can still rename the file and the directory. I think we can simply having the `rename (Directory | File)` for the rename case.

The reason why I try to distinguish them is because I worked in another project written in C for Windows files handling and I remember some bugs are cause by mix-up the directory and file caused so much trouble. So, I just want to place safe and set the `FILE_DIRECTORY_FILE` flag for directory and `FILE_NON_DIRECTORY_FILE` flag for file when opening them (So that the server will return NTSTATUS `STATUS_NOT_A_DIRECTORY` or `STATUS_FILE_IS_A_DIRECTORY`). But seems this not applied to SMB2 rename case by my testing result. So, I think we can ignore the differences for the rename case.

I would combine the functions as `rename(Directory | File)`.

src/main/java/com/hierynomus/smbj/share/DiskShare.java
```diff
@@ -466,6 +468,127 @@ private boolean exists(String path, EnumSet<SMB2CreateOptions> createOptions) th
...
     }
 
+    /***
+     * Rename(Move) Directory/File. Default will using the false for replaceIfExist option
+     * @param path the path to the original directory/file
+     * @param newFileName the new directory/file name/path after rename
+     * @throws TransportException
+     * @throws SMBApiException
+     */
+    public void rename(String path, String newFileName) throws TransportException, SMBApiException {
```
hierynomus: Ahh you misunderstood me. We have both a `File` and a `Directory` class which are returned from the `getFile` call in `DiskShare`. Why did you opt for putting the `rename` operation on the `DiskShare`, instead of implementing it in the `File` and `Directory` classes :)

yin19941005: Ok lol, I got what you mean now. Actually I didn't think too deep on why I putting it on the DiskShare but these are the reasons:

1. Able to reuse the setInfoCommon 
I want to implement for the set methods for `FileBasicInformation` at the beginning. I found the only usage for `SMB2QueryInfoRequest` and the `queryInfoCommon` in the DiskShare. I simply think the getter and setter should in the same class and I think I better create a method for setInfoCommon, just like the `queryInfoCommon` and can reuse for those set methods. After that, I just simply think the rename operation can also reuse the function so I put it in `DiskShare`.

2.  Convenient to calling it in DiskShare
It doesn't seems convenience to put the it in the `File` and `Directory` classes. In my opinion, usually we won't perform other operations when renaming the file. Also, the rename operation required `DELETE` access right to perform the rename. I think most of the programs will not grant the `DELETE` access right by default. That means if we put it in `File` and `Directory` class, we usually got somethings like this : 
```
file.close()
file = share.openFile(... , toLong(EnumSet.of(GENERIC_WRITE, DELETE)), ...)
file.rename(newpath)
file.close()
```
Of course, we can simply create a `mv` method for the `DiskShare` and call the `file.rename` to solve this problem, but I just think most of the use case will just simply calling the `DiskShare.mv()` / `DiskShare.rename()` method so I just simply put it in `DiskShare`.

hierynomus: Ok, The File/Disk APIs are still hardly designed unfortunately, but what I would prefer is to have a `rename` (and/or `mv`) method on both `Disk` and `File` which checks whether the Entry was opened with the correct access level, and executes the command. Failing if the `DELETE` access has not been given in the `getFile`, `openDirectory` or `openFile` call.

API-wise it is nicer to have them on those domain objects, at least that is my opinion :)

And yes maybe we should move some more methods also while we're at it. I've been meaning to refactor that part anyway.

yin19941005: Ok, actually I agree putting the `rename` on both `Directory` and `File` is one of the good designs. If a bug or exception occur only with directory but not file, we can easier to modify and fix it. I could made another pull request for the rename implement on both `File` and `Directory`.

> And yes maybe we should move some more methods also while we're at it. I've been meaning to refactor that part anyway.

I am not sure I got what you mean, do you mean you are going move those `getFileInformation` methods to `File` and `Directory` and add mpre methods to them?

hierynomus:
You can also update this PR :)

yin19941005:
Because of #59 and the suggestion given above, reworked #61 for the rename function.

PR #53, opened 2017-06-06
Use custom FileTime instead of Date to avoid losing precision

pepijnve:
Date only provides millisecond granularity while SMB (and most backing servers) support nanosecond granularity. Using a custom FileTime class avoids loss of information in the protocol implementation. This allows timestamps to be transferred between servers without loss of precision.

### DIFF
--- a/src/main/java/com/hierynomus/msdtyp/FileTime.java
+++ b/src/main/java/com/hierynomus/msdtyp/FileTime.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright (C)2016 - SMBJ Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


src/main/java/com/hierynomus/msdtyp/FileTime.java
```diff
@@ -0,0 +1,59 @@
+package com.hierynomus.msdtyp;
```
hierynomus: If you run `gradle licenseFormat` it will add the license header to this file, which should make the build green.

src/main/java/com/hierynomus/msdtyp/FileTime.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public static FileTime now() {
+        return ofEpochMillis(System.currentTimeMillis());
+    }
+
+    public static FileTime ofEpochMillis(long epochMillis) {
+        return new FileTime(epochMillis * NANO100_TO_MILLI + WINDOWS_TO_UNIX_EPOCH);
+    }
+
+    public FileTime(long windowsTimeStamp)
```
hierynomus: Can you ensure it's consistently formatted? I.e. some methods have open curly at end of line (in line with current code style) and some have it on a new line 😉

hierynomus:
Thanks for the PR. It indeed makes sense to do that. If you can fix the comments (and the build is green), I'll merge it.

hierynomus:
Thanks again!

PR #54, opened 2017-06-06
Added support for iterable Directory and Removed SMB2_INDEX_SPECIFIED.

sopak:
One big list is not usable for large directories.

SMB2_INDEX_SPECIFIED is not widely supported and cause problems for example in NetApp 7mode.
As we can say it is ignored, but some file servers have problems when it is set up.

### DIFF
--- a/src/main/java/com/hierynomus/smbj/share/Directory.java
+++ b/src/main/java/com/hierynomus/smbj/share/Directory.java
@@ -172,9 +172,8 @@ public class Directory extends DiskEntry implements Iterable<FileInfo> {
                     FileInfo next = ctx.next;
                     ctx.iteratorIndex++;
                     return next;
-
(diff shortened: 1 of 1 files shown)


src/main/java/com/hierynomus/smbj/share/Directory.java
```diff
@@ -105,4 +126,57 @@ public String toString() {
...
+            @Override
+            public FileInfo next() {
+                try {
+                    queryIfNeeded();
+                    FileInfo next = ctx.next;
+                    ctx.iteratorIndex++;
+                    return next;
+
+                } catch (TransportException e) {
+                    return null;
```
hierynomus: Not sure if it's the best thing here to return `null`.

hierynomus:
Thanks again :)

PR #55, opened 2017-06-07
Updated iterable directory

sopak:
These changes are based on patch from @pepijnve , cleaner solution of an iterator.

### DIFF
--- a/src/main/java/com/hierynomus/smbj/share/Directory.java
+++ b/src/main/java/com/hierynomus/smbj/share/Directory.java
@@ -172,9 +172,8 @@ public class Directory extends DiskEntry implements Iterable<FileInfo> {
                     FileInfo next = ctx.next;
                     ctx.iteratorIndex++;
                     return next;
-
                 } catch (TransportException e) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/hierynomus/smbj/share/Directory.java
```diff
@@ -126,56 +73,87 @@ public String toString() {
...
-                    return next;
-                } catch (TransportException e) {
-                    throw new SMBRuntimeException(e);
+            NtStatus status = qdResp.getHeader().getStatus();
+
+            if (status == NtStatus.STATUS_NO_MORE_FILES) {
+                return null;
+            } else {
+                if (status != NtStatus.STATUS_SUCCESS) {
+                    throw new SMBApiException(status, qdResp.getHeader().getStatusCode(), SMB2MessageCommandCode.SMB2_QUERY_DIRECTORY,
```
hierynomus: Prefer to use the `SMBApiException(SMB2Header header, String message)` constructor. That one should work here :)

PR #56, opened 2017-06-07
Modify converting flexibility with TimeUnit

sopak:


src/main/java/com/hierynomus/msdtyp/FileTime.java
```diff
@@ -16,6 +16,7 @@
 package com.hierynomus.msdtyp;
 
 import java.util.Date;
+import java.util.concurrent.TimeUnit;
 
 import static com.hierynomus.msdtyp.MsDataTypes.NANO100_TO_MILLI;
```
hierynomus: Do we still need this constant?

sopak: Maybe we should move all of them into FileTime

hierynomus: Yes, that makes sense now that that's modeled out as a class :)

PR #58, opened 2017-06-07
Setting FileId from DiskShare.getFile call (Fixes #57)

### LINKED ISSUE #57: DiskShare.getFile does not set fileId on returned File
Hi folks,

I just started digging into smbj and the whole smb topic on the technical level and encountered a behaviour that I would not have expected when it came down to using an input stream.

So I connected to a share and retrieved a File object via the getFile method of the DiskShare class. Unfortunately this File object could not be used to create an input stream as it would fail with a NullPointerException when calling read on the input stream.

If I get it correctly it seems that the fileId of the underlying File object was not set by getFile method although the fileId was available in the retrieved FileInfo. Is there a reason that the fileId from the FileInfo was not used in the created File object or is it just missing?

Regards,

Pernicies

hierynomus:
Just need to fix the integration test to prove it works.


hierynomus:
Superseded by #71

PR #59, opened 2017-06-07
Make FileInformation querying more fine grained

pepijnve:
This PR splits FileInfo into its constituent parts. This in turn allows finer grained info requests to be performed rather than always querying FileAllInformation and FileIdBothDirectoryInformation.
I've added some preparation work to enable setInfo calls as well for those information classes that are settable.


pepijnve:
@hierynomus Better Code Hub is complaining but I can't see the results (owner and collaborators only it says). If you could tell me what's wrong I'll fix it right away.

hierynomus:
@pepijnve BCH I think I'm going to disable. Though some of their guidelines make sense, not all of them do for a low-level library. I do care about the codacy stuff :) Please have a look at that because that also flagged the PR.

hierynomus:
Had a quick look at the PR. Do we need all the equals/hashcode methods? I don't think we're ever storing them in a collection or comparing them.

pepijnve:
re: equals/hashCode, I was thinking of using those in unit tests but didn't in the end. Would you prefer to see them gone?

hierynomus:
Yes, let's get rid of the equals/hashcodes.

The long parameter lists will need to take a look at. It's one of the symptoms of (semi-)immutable objects that everything is injected in the constructors ;)

pepijnve:
Alternative is a Builder, but that seems needlessly complex for this particular case. Or drop immutability and set each field directly. The latter might actually be worthwhile for listing huge directories. Rather than creating a distinct File...Information object per entry you could reuse the same one. Bit of a pitfall for API users though...

hierynomus:
Yes, that is a trap I've fallen into once or twice myself. .. Why does this object keep changing on each next call... :-/ Ohhh, the library's being smart.

The good thing is that the Directory/FileInfo iterators are lazy, so iterating a huuuuge directory should not blow up the memory on its own. Maybe put a warning on top of the `list` method that it eagerly loads all entries.

pepijnve:
All checks pass now, so I would say we leave things as is for now and not try to be clever. Get stuff working correctly first :)

pepijnve:
The codacy checks only showed up after some time. Still complaining about the long parameter lists...
I've added `@SuppressWarnings("PMD.ExcessiveParameterList")` annotations to the offending constructors. Is that ok with you for these cases @hierynomus ? IMO solving this to please PMD will make the code more complex rather than simpler.

PR #61, opened 2017-06-08
Re-Implement rename function in file and directory

yin19941005:
Using the new FileInformation classes and the generic encoder to implement the rename function. 

Created the FileRenameInformation class.

PR #63, opened 2017-06-08
Added remote rename support

oranoran:
Hi!
I added support for remote rename. This required fixing the UTF including name to not include the byte order mark.
While writing the rename() method, I refactored some common code that gets a file handle and then calls setInfoCommon() on it, into a new method called createAndSetInfoCommon().


hierynomus:
@oranoran Thanks for the PR. But I also have #61 as a PR for renames open. If the codacy warnings for that are fixed, I'll be merging that one in favour of this I think.

oranoran:
Ok, I fixed my own indentation issues anyway so you still have the option of using this PR.

hierynomus:
Merged #61 for this.

PR #64, opened 2017-06-08
Backup intent support

sopak:
Added ability to set createOptions on Share class.

In some cases, you need to setup SMB2CreateOptions.FILE_OPEN_FOR_BACKUP_INTENT
on every open file. Mainly for restore purposes.

More info is here https://msdn.microsoft.com/en-us/library/cc246502.aspx.


pepijnve:
@sopak This one doesn't make sense to me. Why add this mechanism when you already have a 'create options' parameter on `open`?

sopak:
@pepijnve it is not propagated everywhere, and let say it emulate mount with root.
You can use all methods without rewriting the internal application logic but with gained privileges.

pepijnve:
Up to @hierynomus, but I wouldn't do this. Too easy to misuse. You could for instance add `FILE_DIRECTORY_FILE` at the share level and get all kinds of weird behaviour. Maybe a bit far fetched, but it illustrates my point.
From an API design pov I would be inclined to mimic the Win32 API as much as possible. I.e., no create options at the share level, only as `open`/`create` parameters.

sopak:
Ok, I can reduce it to specific values.

My opinion is FILE_OPEN_FOR_BACKUP_INTENT is on open/create by accident, and it never should be there.
It makes perfect sense to place it when you are connecting to the share.

hierynomus:
@sopak I'm also not in favour of this. If you want to do this, you can very easily wrap the DiskShare in a decorator that adds the Create Options on the `open` calls.

What I will add to ease this use-case is a Visitor pattern to ease iterating over directories/files.

PR #66, opened 2017-06-09
Public the constructor in FileInformation Classes

yin19941005:
Base on my understanding, I think the constructor of these FileInformation Classes should be `public`. Minor fix for them.


pepijnve:
Only the settable ones need to be public. There's not much point in instantiating a query only information object directly. There's no harm in their constructor being public, but there's not much you can do with it either.

hierynomus:
Agreed with @pepijnve, only make the 'settable' ones public, keep the rest package private.

yin19941005:
Ok, I see your point. Actually only the `FilePositionInformation` is 'settable' in the classes which I changed. I will create another pull request to just make it public.

PR #67, opened 2017-06-09
Allow set info FileDispositionInformation

sopak:
Add ability to set DeleteOnClose on already opened file.

### DIFF
--- a/src/main/java/com/hierynomus/msfscc/fileinformation/FileInformationFactory.java
+++ b/src/main/java/com/hierynomus/msfscc/fileinformation/FileInformationFactory.java
@@ -264,6 +264,20 @@ public class FileInformationFactory {
                 return parseFileNamesInformation(inputBuffer);
             }
         });
+
+        FileInformation.Encoder<FileRenameInformation> renameCodec = new FileInformation.Encoder<FileRenameInformation>() {
+            @Override
+            public FileInformationClass getInformationClass() {
+                return FileInformationClass.FileRenameInformation;
+            }
+
+            @Override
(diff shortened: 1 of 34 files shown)


src/main/java/com/hierynomus/msfscc/fileinformation/FileInformationFactory.java
```diff
@@ -107,6 +107,25 @@ public void write(FileBasicInformation info, Buffer outputBuffer) {
...
+            }
+
+            @Override
+            public FileDispositionInformation read(Buffer inputBuffer) throws Buffer.BufferException {
+                throw new UnsupportedOperationException();
+            }
+
+            @Override
+            public void write(FileDispositionInformation info, Buffer outputBuffer) {
+                outputBuffer.putRawBytes(FileInformationFactory.getFileDispositionInfo(true));
```
hierynomus: Shouldn't the boolean option be encoded in the `FileDispositionInformation` class?

sopak: Do you mean set it also to false?
As if one operation will delete the file and another will discard it on the same opened fileId?

hierynomus: No, using `false` you can also unmark it for deletion. See: http://download.microsoft.com/download/4/3/8/43889780-8d45-4b2e-9d3a-c696a890309f/File%20System%20Behavior%20Overview.pdf

sopak:
I did something wrong here. I will create a new pull request.

PR #68, opened 2017-06-09
Add setSecurityInfo

pepijnve:
This PR adds DiskShare#setSecurityInfo
I've reviewed the binary security descriptor marshalling/unmarshalling code and tested that against a couple of servers.
I also added SDDL parsing and formatting code. Both are incomplete in the sense that they don't support conditional ACEs and resource attribute ACEs yet, but those are not used all that often for file system objects (as far as I know).


pepijnve:
I've resolved the PMD warnings in the handwritten parts of the SDDL parser.
The other warnings were in generated parser code. I've added `@SuppressWarnings("PMD")` to those classes since it's rather pointless to try and fix those.

hierynomus:
@pepijnve, I'm doubting about this PR. There is a boatload of generated code in there that I really don't like. Also I think we could take some of the refactorings/renamings of the ACE/Acl and make them a separate PR.

As for the parser, is it possible to generate the code as part of the build using that parser/generator?

pepijnve:
> Also I think we could take some of the refactorings/renamings of the ACE/Acl and make them a separate PR.

Sure, easy enough to split it up into smaller parts. I'll make the SDDL parser/formatter a distinct PR as well.

> As for the parser, is it possible to generate the code as part of the build using that parser/generator?

Sure, but then you do have to take the generated classes as-is. That means they're all public. As an alternative the default runtime package could be used, but then you get a dependency on an external runtime library.

pepijnve:
One potential speed bump is that the latest version of Mouse on repo1 is pretty outdated (http://repo1.maven.org/maven2/se/romanredz/mouse/mouse-runtime/1.5.1/). Any suggestions on how to deal with that?

pepijnve:
@hierynomus FYI I'm going to rework this when (if) PR #71 is completed. Saves me from doing lots of rebasing work.

pepijnve:
I created a smaller version of this PR that only adds setSecurityInformation methods as PR #73.

Issue #62, opened 2017-06-08
Logging is too verbose


sopak:
I think we should change all occurrences of logger.info to logger.debug. With few exceptions as connect, disconnect and authenticate.

gidravlic:
I support this issue

sopak:
@hierynomus what is your opinion? Can I do this change?

hierynomus:
Merged #83 to make logging a bit less verbose on INFO level. Now you just see the connecting/closing of the different domain objects.

PR #69, opened 2017-06-09
Change verbosity of logging

sopak:
As this library supposed to be low level, I decided
to move all info to the debug level.

Based on the issue #62.

### DIFF
--- a/src/main/java/com/hierynomus/protocol/commons/socket/ProxySocketFactory.java
+++ b/src/main/java/com/hierynomus/protocol/commons/socket/ProxySocketFactory.java
@@ -77,7 +77,7 @@ public class ProxySocketFactory extends SocketFactory {
         if (bindAddress != null) {
             socket.bind(bindAddress);
         }
-        logger.info("Connecting to {}", address);
+        logger.debug("Connecting to {}", address);
         socket.connect(address, connectTimeout);
         return socket;
     }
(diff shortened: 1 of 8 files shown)


hierynomus:
Hi @sopak,

I both agree and disagree 😁 . 

Agreed: Yes, a library is low-level. _But_ that means from a logging perspective that it logs relevant messages for "it's layer" in your application. The nice thing about logging frameworks is that for your application if you are not interested in a certain logger, you can disable it or increase its level.

Agreed: It might be a bit verbose here and there on info level, and we should tweak that. _But_ logging nothing on info level anymore is counterproductive and not in line with the above.

I am willing to merge a PR that goes over the logs with a bit more precision to tune them, but at it currently is, I am not inclined to merge this PR.

sopak:
@hierynomus  ok,  what do you think should be in info, this pull request gives an overview so we can discuss each of them.

pepijnve:
Aren't all the log messages that are in the library at the moment debug level logging, i.e. messages that are only useful when analysing the behaviour of the library itself?
The ones in the DiskShare methods (e.g., `logger.info("exists {}", path);`) should probably just be removed. Seems a bit useless if that were to pop up in log output.

hierynomus:
In the end isn't _all_ logging ever written debug level logging. You use a log when something unexpected happens in your FE and you want to find out (debug) what happened 😉 

Agreed with the `DiskShare` methods, they should be `trace` most probably.

On `info` level I would at least like to see:
- the connection attempt
- the fact it connected
- that it authenticated succesfully
- connect to share
- disconnect

That give you a high level trail of what happened (and when!).

hierynomus:
See #83. Closing this one.

PR #71, opened 2017-06-11
Make file/dir operations available as instance methods on File/Directory

### LINKED ISSUE #12: Open a FileId once, and re-use that for multiple operations on the file (existence, read, write, fileinfo, etc)

### LINKED ISSUE #57: DiskShare.getFile does not set fileId on returned File

pepijnve:
Fixes #12 by making file operations from DiskShare available as instance methods on DiskShare/File/Directory.
Also fixes #57.

### DIFF
--- a/src/main/java/com/hierynomus/mssmb2/messages/SMB2Flush.java
+++ b/src/main/java/com/hierynomus/mssmb2/messages/SMB2Flush.java
@@ -32,8 +32,9 @@ public class SMB2Flush extends SMB2Packet {
         super();
     }
 
-    public SMB2Flush(SMB2Dialect smbDialect) {
+    public SMB2Flush(SMB2Dialect smbDialect, SMB2FileId fileId) {
         super(24, smbDialect, SMB2MessageCommandCode.SMB2_FLUSH);
+        this.fileId = fileId;
     }
 
     @Override
@@ -48,9 +49,5 @@ public class SMB2Flush extends SMB2Packet {
     protected void readMessage(SMBBuffer buffer) throws Buffer.BufferException {
         buffer.readUInt16(); // StructureSize (2 bytes)
         buffer.skip(2); // Reserved (2 bytes)
-    }
-
-    public void setFileId(SMB2FileId fileId) {
-        this.fileId = fileId;
     }
 }
--- a/src/main/java/com/hierynomus/smbj/session/Session.java
+++ b/src/main/java/com/hierynomus/smbj/session/Session.java
@@ -78,10 +78,10 @@ public class Session implements AutoCloseable {
      * @return the handle to the connected share.
      */
     public Share connectShare(String shareName) {
-        TreeConnect connectedConnect = treeConnectTable.getTreeConnect(shareName);
-        if (connectedConnect != null) {
-            logger.info("Returning cached Share {} for {}", connectedConnect.getTreeId(), shareName);
-            return connectedConnect.getHandle();
+        Share connectedShare = treeConnectTable.getTreeConnect(shareName);
+        if (connectedShare != null) {
+            logger.info("Returning cached Share {} for {}", connectedShare, shareName);
+            return connectedShare;
         } else {
             return connectTree(shareName);
         }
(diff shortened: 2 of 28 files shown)


src/main/java/com/hierynomus/mssmb2/messages/SMB2Flush.java
```diff
@@ -32,8 +32,9 @@ public SMB2Flush() {
         super();
     }
 
-    public SMB2Flush(SMB2Dialect smbDialect) {
+    public SMB2Flush(SMB2Dialect smbDialect, SMB2FileId fileId) {
         super(24, smbDialect, SMB2MessageCommandCode.SMB2_FLUSH);
+        this.fileId = fileId;
```
hierynomus: Do we need the setter still?

src/main/java/com/hierynomus/smbj/session/Session.java
```diff
@@ -132,11 +136,11 @@ private void disconnectTree(TreeDisconnected disconnectEvent) {
 
     public void logoff() throws TransportException {
         logger.info("Logging off session {} from host {}", sessionId, connection.getRemoteHostname());
-        for (TreeConnect treeConnect : treeConnectTable.getOpenTreeConnects()) {
+        for (Share treeConnect : treeConnectTable.getOpenTreeConnects()) {
```
hierynomus: rename to `share`

src/main/java/com/hierynomus/smbj/share/DiskEntry.java
```diff
@@ -15,101 +15,93 @@
...
+
+    public SecurityDescriptor getSecurityInformation(Set<SecurityInformation> securityInfo) throws SMBApiException {
+        return share.getSecurityInfo(fileId, securityInfo);
+    }
+
+    public void rename(String newName)throws TransportException, SMBApiException {
+        this.rename(newName, false);
+    }
+
+    public void rename(String newName, boolean replaceIfExist)throws TransportException, SMBApiException {
```
hierynomus: formatting, missing space (also below)

src/main/java/com/hierynomus/smbj/share/DiskShare.java
```diff
@@ -105,56 +81,63 @@ public Directory openDirectory(
...
      */
-    public File openFile(String path, EnumSet<AccessMask> accessMask, SMB2CreateDisposition createDisposition) throws TransportException, SMBApiException {
-        logger.info("OpenFile {},{},{}", path, accessMask, createDisposition);
-
-        long accessMaskValue = toLong(accessMask);
-        SMB2FileId fileId = open(path, accessMaskValue, null, EnumSet.of(FILE_SHARE_READ), createDisposition, EnumSet.of(SMB2CreateOptions.FILE_NON_DIRECTORY_FILE));
-        // TODO
-        return new File(fileId, treeConnect, path, accessMaskValue);
+    public File openFile(String path, EnumSet<AccessMask> accessMask, SMB2CreateDisposition createDisposition) throws SMBApiException {
+        return (File) open(path, toLong(accessMask), EnumSet.of(FILE_ATTRIBUTE_DIRECTORY), EnumSet.noneOf(SMB2ShareAccess.class), createDisposition, EnumSet.of(SMB2CreateOptions.FILE_DIRECTORY_FILE));
```
hierynomus: Are the passed options correct? They're the same as for the `openDirectory` call.

src/main/java/com/hierynomus/smbj/share/Share.java
```diff
@@ -40,6 +40,8 @@
 
 public class Share implements AutoCloseable {
     private static final Logger logger = LoggerFactory.getLogger(Share.class);
+    private static final EnumSet<NtStatus> SUCCESS = EnumSet.of(NtStatus.STATUS_SUCCESS);
```
hierynomus: Shouldn't these be `Set` also? ;)

src/main/java/com/hierynomus/mssmb2/messages/SMB2Flush.java
pepijnve: That's up to you. I don't think it's useful anymore, but I left it since it was already there.

src/main/java/com/hierynomus/smbj/share/DiskShare.java
pepijnve: Nope that's wrong. Thanks for catching that.

src/main/java/com/hierynomus/smbj/share/Share.java
pepijnve: 😸 I changed a lot of fields and parameters from EnumSet to Set for added flexibility (for instance to allow using Guava ImmutableSet). Doesn't matter as much here since this is an internal constant. However you prefer.

pepijnve:
Requested changes integrated

Erates:
@hierynomus Any idea when a release will be available for this issue? I would like to rename files, but I can't depend on a non-released version.

Issue #16, opened 2016-07-26
Allow read/write buffer sizes to be configurable


hierynomus:
Now the read buffer size and write buffer size for file read/write is set to the maximum that was negotiated. Allow this to be tweaked.

PR #75, opened 2017-06-12
Make read/write/query buffer sizes configurable (#16)

pepijnve:
Add extra read/write overloads
Consolidate all request creation/parsing in Share


hierynomus:
Can you checkout Codacy?

pepijnve:
Done

hierynomus:
The idea of #16 was to have some Config parameter to tweak the buffer sizes. Does that make sense, or do you think this is more useful to have?

pepijnve:
🤔 you're probably right that having this on a per read/write call level doesn't make much sense.

pepijnve:
Reworked to make buffer sizes configurable via the Config class.

I also reworked Config a bit. Was there a particular reason to have an interface here? Seems like a simple POJO so having an interface for it seemed like overkill. Changed it to a final, immutable class with a Builder and 'create default' factory method.

src/main/java/com/hierynomus/smbj/Config.java
```diff
@@ -15,36 +15,182 @@
...
+        public void setRandomProvider(Random random) {
+            if (random == null) {
+                throw new NullPointerException("Dialect may not be null");
+            }
+            config.random = random;
+        }
+
+        public void setSecurityProvider(SecurityProvider securityProvider) {
+            if (securityProvider == null) {
+                throw new NullPointerException("Dialect may not be null");
```
hierynomus: copy paste error, should be SecurityProvider

src/main/java/com/hierynomus/smbj/Config.java
```diff
@@ -15,36 +15,182 @@
...
+            setRandomProvider(new SecureRandom());
+            setSecurityProvider(new JceSecurityProvider());
+            setReadBufferSize(SMB2Packet.SINGLE_CREDIT_PAYLOAD_SIZE);
+            setWriteBufferSize(SMB2Packet.SINGLE_CREDIT_PAYLOAD_SIZE);
+            setTransactBufferSize(SMB2Packet.SINGLE_CREDIT_PAYLOAD_SIZE);
+        }
+
+        public void setRandomProvider(Random random) {
+            if (random == null) {
+                throw new NullPointerException("Dialect may not be null");
```
hierynomus: copy paste error, should be RandomProvider

src/main/java/com/hierynomus/smbj/Config.java
```diff
@@ -15,36 +15,182 @@
...
+            config = new Config();
+            setClientGuid(UUID.randomUUID());
+            setRandomProvider(new SecureRandom());
+            setSecurityProvider(new JceSecurityProvider());
+            setReadBufferSize(SMB2Packet.SINGLE_CREDIT_PAYLOAD_SIZE);
+            setWriteBufferSize(SMB2Packet.SINGLE_CREDIT_PAYLOAD_SIZE);
+            setTransactBufferSize(SMB2Packet.SINGLE_CREDIT_PAYLOAD_SIZE);
+        }
+
+        public void setRandomProvider(Random random) {
```
hierynomus: If we make it a builder, which I don't oppose to, let's also make the methods `with...` and have them return the builder so you can chain the calls.

src/main/java/com/hierynomus/smbj/share/Directory.java
```diff
@@ -22,6 +22,7 @@
 import com.hierynomus.msfscc.fileinformation.FileInformation;
 import com.hierynomus.msfscc.fileinformation.FileInformationFactory;
 import com.hierynomus.mssmb2.SMB2FileId;
+import com.hierynomus.mssmb2.SMB2Packet;
```
hierynomus: Unnecessary import?

src/main/java/com/hierynomus/smbj/Config.java
```diff
@@ -15,36 +15,182 @@
...
+
+    public static class Builder {
+        private Config config;
+
+        Builder() {
+            config = new Config();
+            setClientGuid(UUID.randomUUID());
+            setRandomProvider(new SecureRandom());
+            setSecurityProvider(new JceSecurityProvider());
+            setReadBufferSize(SMB2Packet.SINGLE_CREDIT_PAYLOAD_SIZE);
```
hierynomus: Maybe have the default be `Integer.MAX_INT`, which means that they'll be taken from the negotiated protocol. Now they'll always be taken from the config, unless you manually set it to MAX_INT (or something higher than negotiated). I think the default should be that it is taken from the negotiation, and can be tweaked from the config.

pepijnve:
Integrated all the changes you requested.

I went for a conservative buffer size to err on the safe side. The buffer size setting also determines how much memory the library is going to allocate for buffers and simply trusting the server seemed risky as a default.
I've added a `withNegotiatedBufferSize` method on the Builder and made that the default as you requested for now.
 Perhaps 1MB as default buffer size would be a better default? That way memory usage is predictable and bounded, but you still use multicredit requests by default.

src/main/java/com/hierynomus/smbj/Config.java
```diff
@@ -15,36 +15,199 @@
...
+    private boolean signingRequired;
+    private SecurityProvider securityProvider;
+    private int readBufferSize;
+    private int writeBufferSize;
+    private int transactBufferSize;
 
-    SecurityProvider getSecurityProvider();
+    public static Config createDefaultConfig() {
+        return builder()
+            .withDialect(SMB2Dialect.SMB_2_1)
```
hierynomus: Add a `withDialects(List<>)` and a `withDialects(Dialect...)`, same for `withAuthenticators`. It makes the Config API more fluent, and also clearer that you can add multiple. Those methods should clear the existing set and replace it with the configured set.

pepijnve:
Third times a charm 😄  Change the multi-valued builder methods to plural with a variadic and Iterable overload.
I also applied the default max buffer size of 1MB change I suggested earlier. So by default you now get the negotiated buffer size with a max of 1MB. This prevents rogue servers from making the client allocate huge buffers in the default configuration.

src/main/java/com/hierynomus/smbj/Config.java
```diff
@@ -15,36 +15,221 @@
...
+
+    public static class Builder {
+        private Config config;
+
+        Builder() {
+            config = new Config();
+            withClientGuid(UUID.randomUUID());
+            withRandomProvider(new SecureRandom());
+            withSecurityProvider(new JceSecurityProvider());
+            withReadBufferSize(SMB2Packet.SINGLE_CREDIT_PAYLOAD_SIZE);
```
hierynomus: These should also be the `DEFAULT_BUFFER_SIZE`

pepijnve:
Doing too many things in parallel. Setting of defaults is now done in one place (`builder()` method). Since you can now set the entire set of dialects and authenticators there's no point in leaving those collections empty by default.

Issue #72, opened 2017-06-12
Data was corrupted at maxWriteSize(8MB) boundary, when File#write() using a buffer of a size not a power of 2.


4dn-oss:
Following code makes corrupted file.

```
OutputStream out = file.getOutputStream();// dst.txt on remote server
FileInputStream in = new FileInputStream("src.txt");// local file larger than 8MB
byte[] buf = new byte[1000];// buffer size is not a 2^n
int n = 0;
while ((n = in.read(buf)) != -1) {
	out.write(buf, 0, n);
}
out.flush();
out.close();
```

```
# cmp src.txt dst.txt
src.txt dst.txt differ: char 8388609, line 87382
```

Java: Java(TM) SE Runtime Environment (build 1.8.0_112-b15)
Server: Windows 10 Pro (negotiatedProtocol=NegotiatedProtocol{dialect=SMB_2_1, maxTransactSize=8388608, maxReadSize=8388608, maxWriteSize=8388608})

It seems to be a bug of RingBuffer.

```java:src/main/java/com/hierynomus/smbj/share/RingBuffer.java
     private void writeBytes(byte[] b, int off, int len) {
         if (writeIndex + len <= buf.length) {
             System.arraycopy(b, off, buf, writeIndex, len);
         } else {
             int bytesToEnd = buf.length - writeIndex;
             System.arraycopy(b, off, buf, writeIndex, bytesToEnd);
-            System.arraycopy(b, bytesToEnd, buf, bytesToEnd, len - bytesToEnd);
+            System.arraycopy(b, bytesToEnd, buf, 0, len - bytesToEnd);
         }
     }
```

PR #76, opened 2017-06-12
Correct RingBuffer write logic when wraparound occurs (#72)

pepijnve:


hierynomus:
Thanks, I wanted to look at that tonight :) You make my life a whole lot easier this way!

Issue #80, opened 2017-06-14
DiskShare.getFileInformation() does not work properly for root directory


swtet2003:
We built a version of SMBJ when this was the head of master:

https://github.com/hierynomus/smbj/commit/57dfc9825b6d0f4935cf787e4fba8ac1ce7e9632

With that code, the following code snippet would report inconsistent results between Windows 10 and Windows Server 2012.  For Win 10, the name information was "".  For 2012, it was "\temp".  Both hosts negotiated SMB_2_1 dialect.

DiskShare share = (DiskShare)session.connectShare( "temp" );
FileAllInformation file = share.getFileInformation( "" );
System.out.println( "File info name: " + file.getNameInformation() );

I just rebuilt today what is in master (hoping that the issue had already been resolved) and now using "" or even null does not work.  I get the following error, using "":

com.hierynomus.smbj.common.SMBApiException: STATUS_SHARING_VIOLATION(3221225539/3221225539): Create failed for
        at com.hierynomus.smbj.share.Share.receive(Share.java:219)
        at com.hierynomus.smbj.share.Share.sendReceive(Share.java:198)
        at com.hierynomus.smbj.share.Share.createFile(Share.java:115)
        at com.hierynomus.smbj.share.DiskShare.open(DiskShare.java:64)
        at com.hierynomus.smbj.share.DiskShare.open(DiskShare.java:60)
        at com.hierynomus.smbj.share.DiskShare.getFileInformation(DiskShare.java:159)
        at com.hierynomus.smbj.share.DiskShare.getFileInformation(DiskShare.java:152)

Any idea how I can resolve either of these issues?

pepijnve:
This is due to the code path that accepts a string path not specifying any share modes. It used to specify all in earlier versions. I'll fix that asap.
If you need a short term workaround, you can fix this by using `openDirectory` with access mask `GENERIC_READ`, a set of all the ShareAccess values and create disposition `FILE_OPEN`. Then call `getFileInformation` on the returned Directory object followed by `close`.

hierynomus:
Fixed with merging #81

swtet2003:
Any chance you can update SMBTest to reflect the API changes?

With the changes that were merged to master, I am back to my original problem where Win 10 and 2012 report different information using DiskShare.getFileInformation().  Any idea how to fix that?

pepijnve:
[...]

PR #81, opened 2017-06-15
Use SMB2ShareAccess.ALL in convenience methods.

pepijnve:
This fixes issue #80.
In the process I removed the limited variants of openDirectory/File. These are rarely going to be useful since the create options and create disposition are rather critical parameters. The alternative would be to introduce special cases like openExistingDirectory, createNewDirectory, etc but that would lead to a multitude of special case methods.


hierynomus:
Codacy has some problems due to unneeded qualified imports ;)

pepijnve:
Ugh, that's why I don't use these tools. They're way too pedantic. I'll get these fixed asap...

hierynomus:
I like the fact that it enforces a more or less consistent style at least in the file. But IDEs could be better here also when they change to `*` import, to static import all of the occurrences ;)

PR #84, opened 2017-06-15
Expose QueryDirectory search pattern

pepijnve:


pepijnve:
Just FYI, while testing I ran into an issue running queries against macOS. With a search pattern that matches a single file the server never sends back a STATUS_NO_MORE_FILES response back. It just keeps sending the same response with a single result back.
If I use a pattern with a wildcard that also only matches a single entry, the proper STATUS_NO_MORE_FILES response is sent back.

hierynomus:
@pepijnve Any response from apple on this?

Is there some way to detect this corner case? Compare the last entry to the next one up, and have the iterator stop if they're the same?

pepijnve:
No response and I wouldn't expect one soon. I was thinking of adding a heuristic like the one you're proposing indeed. I don't see any other way around this.

hierynomus:
Yes, I think adding that, and adding some comments there to explain why this is needed would be the best thing.

swtet2003:
Even though this hasn't been accepted yet, I thought I would pull it and see if it fixed my issue.  It does not appear to make any difference.  Should I expect DiskShare.getFileInformation() to return something different now based on these changes?  Or, do I need to do something differently?

pepijnve:
No this doesn't affect `getFileInformation`. What it does enable is obtaining the actual name by doing a directory listing with a search pattern that is the file name you're looking for.

swtet2003:
Why would you leave getFileInformation() broken?  Doesn't it make sense to fix it too?

pepijnve:
Because IMO it's not broken. It's correctly sending the request to the server, correctly parsing the response and passing it on to the caller.

pepijnve:
@hierynomus added a generic workaround (response identical to previous response) for the macOS issue.

swtet2003:
I understand that it is technically correct but, at some point, shouldn't the library help smooth out the Microsoft quirks and give the user something consistent, when it can?  If 5,000 people use SMBJ, why should 5,000 people be responsible for coding around SMB corner cases when SMBJ could handle it?

pepijnve:
The main issue is that there is no way to correct the response from the server. What should SMBJ fill in instead of the server's response? What criterium do you use to consider a response incorrect and needing correction? Etc.
You would also need to test this against the various SMB servers out there for correctness. Is the logic correct for Samba, macOS, NetApp, EMC, Hitachi, ...

swtet2003:
Agree with all of that....we have seen plenty of crappy SMB server implementations over the years that have their own little quirks.  I just think the library is the place to deal with it.

pepijnve:
Fair enough. I'm not the maintauner of this library though so final call is with @hieronymous I think. If you want this fixed I think it's best you make a PR with the proposed fix. It's not something I'm willing to spend time on myself I'm afraid since I don't agree with the change.

hierynomus:
Unless you can come up with a very good solution for how to fix the fact that the server omits information, I would also not be inclined to add something like that.

What I can imagine is adding a quirks/compatibility mode which you could set in the config. For instance `config.setQuirksMode(SmbQuirks.Win10)`, but then what should it do in that case for this specific problem? It's not like it can come up with a filename out of thin air....

hierynomus:
I'll reopen #80 to discuss further there.

src/main/java/com/hierynomus/smbj/share/Directory.java
```diff
@@ -38,8 +38,12 @@
     }
 
     public <F extends FileDirectoryQueryableInformation> List<F> list(Class<F> informationClass) throws SMBApiException {
+        return list(informationClass, null);
+    }
+
+    public <F extends FileDirectoryQueryableInformation> List<F> list(Class<F> informationClass, String searchPattern) {
```
hierynomus: Maybe add a bit of JavaDoc to the method to define what can go into a `searchPattern`

src/main/java/com/hierynomus/smbj/share/Directory.java
```diff
@@ -52,7 +56,11 @@
     }
 
     public <F extends FileDirectoryQueryableInformation> Iterator<F> iterator(Class<F> informationClass) {
-        return new DirectoryIterator<>(informationClass);
+        return iterator(informationClass, null);
+    }
+
+    public <F extends FileDirectoryQueryableInformation> Iterator<F> iterator(Class<F> informationClass, String searchPattern) {
```
hierynomus: See above.

pepijnve:
Added javadoc. The actual detailed documentation is on `Directory#iterator(Class,String)`. All the other methods have hyperlinks that eventually end up there.

Note that while reviewing the documentation and code I stumbled upon the fact that `.` and `..` are removed from the result. In the spirit of returning what the server provides unmodified, I've removed that bit of code.

PR #86, opened 2017-06-16
Remove all indefinite waits for responses

### LINKED ISSUE #85: No timeout while waiting on SMB2TreeDisconnect and SMB2Logoff responses
I've run into a case where for some reason no response was received from the SMB server to a SMB2TreeDisconnect request.
As a result, my calling thread hung. Also a PacketReader thread probably remains open in this case (haven't verified that).

I think there should definitely be a timeout in the Future.get() call that is waiting for the response.
Not sure what kind of configuration / API changes should govern this - any suggestions?

BTW the same thing seems to happen during Session.close() with waiting for the SMB2Logoff response, should probably add a timeout there too.

pepijnve:
Fixes #85 by adding configurable timeouts on all operations that wait for responses

### DIFF
--- a/src/main/java/com/hierynomus/smbj/Config.java
+++ b/src/main/java/com/hierynomus/smbj/Config.java
@@ -25,9 +25,13 @@ import com.hierynomus.smbj.auth.SpnegoAuthenticator;
 
 import java.security.SecureRandom;
 import java.util.*;
+import java.util.concurrent.TimeUnit;
 
 public final class Config {
     private static final int DEFAULT_BUFFER_SIZE = 1024 * 1024;
+
+    private static final int DEFAULT_TIMEOUT = 60;
+    private static final TimeUnit DEFAULT_TIMEOUT_UNIT = TimeUnit.SECONDS;
 
     private Set<SMB2Dialect> dialects;
     private List<Factory.Named<Authenticator>> authenticators;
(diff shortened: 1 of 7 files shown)


oranoran:
PR #86 makes sure there are no hang situations, but the unwinding process in case of a server failure may still take a long time.

A typical error handling sequence in case of timeout would make a close() request on the Session, which in turn triggers close() requests on individual shares. In case the server is not responding, each one of these requests would wait the full "transact" timeout period, so the full recovery process can take a few minutes, blocking the calling thread for the duration.

A possible set of improvements would be:

1. TreeConnect.close() could use a separate "close" timeout value, that would be by default much shorter than the "transact" value.
2. Session.close() could wait on all the Future objects of the various shares together, rather than one at a time. This requires changing the Share interface to expose the needed Future objects.
3. Session.close() doesn't really need to wait for the server's response to its own logoff request. Or - it can also wait wait for the shorter "close" timeout period.

Thoughts?

hierynomus:
Is this one superseeded @pepijnve ? I was about to merge it to find it gone now....

pepijnve:
Oh dear. Did I just delete the wrong branch? 😟

hierynomus:
You might've ;) Luckily github still has the history..

pepijnve:
I did. The git GC hadn't run yet so I'm restoring the commit. I'll rebase to your Config -> SmbConfig changes before reopening this one.

pepijnve:
It doesn't look like I can reopen this PR so I'll create a new one.

PR #90, opened 2017-06-19
NamedPipe is now completely not usable - make it usable at least in a low-level way

oranoran:
In a recent code change, the open() method of Share has become protected. As a result, there is absolutely nothing that can be done with a NamedPipe object.
This breaks existing code that was built on top of NamedPipe, using Share.open().

Eventually, there sould be an open() method in NamedPipe that would return a proper remote object, similar to DiskShare.open(). But until a full implementation of a named pipe is available, there needs to be at least a low-level way of opening a handle on it.

So I added an openFileId() method to NamedPipe, which simply delegates to the super class.

### DIFF
--- a/src/main/java/com/hierynomus/smbj/share/NamedPipe.java
+++ b/src/main/java/com/hierynomus/smbj/share/NamedPipe.java
@@ -15,10 +15,22 @@
  */
 package com.hierynomus.smbj.share;
 
+import java.util.Set;
+
+import com.hierynomus.msdtyp.AccessMask;
+import com.hierynomus.msfscc.FileAttributes;
+import com.hierynomus.mssmb2.SMB2CreateDisposition;
+import com.hierynomus.mssmb2.SMB2CreateOptions;
+import com.hierynomus.mssmb2.SMB2FileId;
+import com.hierynomus.mssmb2.SMB2ShareAccess;
 import com.hierynomus.smbj.common.SmbPath;
 
 public class NamedPipe extends Share {
     public NamedPipe(SmbPath smbPath, TreeConnect treeConnect) {
         super(smbPath, treeConnect);
+    }
+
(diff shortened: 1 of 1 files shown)


hierynomus:
@oranoran Did you try `Session.connectTree(String)`? This should return a `NamedPipe` if it detected it is a pipe.

hierynomus:
Sorry, I meant `Session.connectShare(String)`, `connectTree` is the private method underneath...

pepijnve:
@hierynomus I don't think that will work. I only started investigating how pipes over SMB work exactly. If I understand it correctly you need to do a tree connect to the `IPC$` share, which will be of type pipe and then open files for the named pipe you want to open. I don't think you can't connect to the pipe directly.

hierynomus:
Yes, so I think calling `Session.connectTree("IPC$")` would actually return you a `NamedPipe` that is connected. The NamedPipe probably functions like a DiskShare in some way. I also have not looked closely at how it exactly works, also mainly as we needed the File/Directory stuff first ;)

pepijnve:
So the issue, and what this PR is for, is that NamedPipe doesn't actually have any public methods anymore. So once you have that NamedPipe, there's not much you can do with it.

At the very least we need to be able to get a file handle so that you can read/write from/to a pipe. But better would be have a method like `Pipe openPipe(String path, ...)`. Pipe should then have read/write methods and wrappers for the various `FSCTL_PIPE_...` IOCTLs.

hierynomus:
Ahh yes, I hadn't looked at the code change yet. I could merge this as a temporary fix, but it will definitely not be the final API :) If you're OK with that @oranoran it's good for me.

oranoran:
Sure @hierynomus that's the idea.

oranoran:
Actually, as I'm using this fix myself, turns out the same thing needs to be done for the closeFileId(). Will update the PR in a sec.

hierynomus:
OK, will merge this for temporary fix

Issue #91, opened 2017-06-19
Ability to fine-tune socket options (timeout, keepalive, linger, etc.)


oranoran:
It would be useful to be able to set socket options for the sockets used by SMB connections.
Specifically, SO_TIMEOUT, and probably KEEPALIVE and SO_LINGER seem to be useful.

I created a [potential fix](https://github.com/intigua/smbj/tree/socket-options), allowing Config to receive a customized SocketFactory, and allowing ProxySocketFactory to be customized with some socket options.

The only problem is that when I'm testing this fix with the following code, even though sockets get set with the desired socket options, their behavior doesn't change at all. Any ideas?

```
@Test
public void checkFactorySocketTimeout() throws InterruptedException, IOException {
    ProxySocketFactory factory = new ProxySocketFactory();
    factory.setSoTimeout(100);
    factory.setKeepAlive(true);
    Config config = Config.builder().withSocketFactory(factory).build();
    Connection connection = new Connection(config, new SMBEventBus());
    connection.connect(myhost);
    Thread.sleep(200);
    assertFalse(connection.isConnected());
}
```

hierynomus:
There is no need to inject/change `SocketFactory`. The `SocketClient` actually takes care of setting `soTimeout`, and should take care of the rest also. Adding those as config parameters makes sense indeed, I was thinking the same when Pepijn sent the PR this weekend.

oranoran:
Good catch, somehow I missed it.
And unlike my change, this works, though I'm not sure what the difference is.

But adding connection.setSotTimeout(100) to my test still fails the test, since connection.isConnected() continues to return true. The timeout event kills the PacketReader thread, but it doesn't cause the socket to be closed Wouldn't it make sense to close the socket and set it to null in this case, since there is now no thread that can read it?

hierynomus:
Yes, I think that makes sense.

oranoran:
Ok, I will try to create a PR for this (hopefully tomorrow). Probably should also verify that an SO timeout causes Future.get() requests to fail.

hierynomus:
It wouldn't directly fail the Futures, as only when a Packet is received we know for which Future it is. It will however close complete all the waiting futures if we close the connection, due to the SO Timeout.

oranoran:
About using setSoTimeout - I'm not sure whether calling this method will have an effect if called after the socket is already connected. In case it doesn't have an effect, then the only way to effectively set the timeout is not to use SMBClient and create connections directly.

[...]

PR #93, opened 2017-06-20
Config and SMBClient support for socket timeout

oranoran:
Following discussion in #91 
SMBClient now sets the socket timeout when creating a connection.
Otherwise, for any connection created by SMBClient, it's already too late to set sotimeout (their first blocking operation is already under way).


hierynomus:
I just fixed that a moment ago, with the additional cleanup in fee525c

oranoran:
Cool :-)

PR #95, opened 2017-06-20
Made it possible to write byte[] larger then maxWriteBufferSize.

4dn-oss:
Add some process not to raise 'RingBuffer of length xxxx cannot accomodate XXXX bytes.' exception.
Fix RingBuffer#write() logic when wraparound occurs with non-zero offset.


pepijnve:
Might be good to add a test case in RingBufferTest for the non-zero offset write.

hierynomus:
Can you submit one or more testcases for this scenario?

4dn-oss:
OK, I'll try it.

src/test/groovy/com/hierynomus/smbj/share/RingBufferTest.groovy
```diff
@@ -232,6 +232,24 @@ public class RingBufferTest extends Specification {
     byteArray == [5, 6, 7, 8, 9, 10] as byte[]
   }
 
+  def "should be able to append bytes in the middle of array if write position has wrapped around"() {
+	given:
```
hierynomus: Can you apply the right formatting? Looks good otherwise.

PR #99, opened 2017-06-21
Add PipeShare; add basic methods on NamedPipe

pepijnve:
This PR renames NamedPipe to PipeShare and adds a new class NamedPipe which represents an open named pipe itself.
NamedPipe now has the basic methods available that allow you to work with a pipe over SMB.
I had to rework the IOCTL code a bit. The ControlCode enum was dropped since this was too limiting. In theory any IOCTL can be passed over SMB; not just the ones that have special treatment in the SMB spec.


src/main/java/com/hierynomus/mssmb2/messages/SMB2IoctlResponse.java
```diff
@@ -47,7 +47,7 @@ protected void readMessage(SMBBuffer buffer) throws Buffer.BufferException {
 
         buffer.skip(2); // StructureSize (2 bytes)
         buffer.skip(2); // Reserved (2 bytes)
-        controlCode = valueOf(buffer.readUInt32(), SMB2IoctlRequest.ControlCode.class, null); // CtlCode (4 bytes)
+        controlCode = (int)buffer.readUInt32(); // CtlCode (4 bytes)
```
hierynomus: `readUInt32AsInt`

src/main/java/com/hierynomus/smbj/share/NamedPipe.java
```diff
@@ -15,27 +15,192 @@
...
+
+    public void close() {
+        share.closeFileId(fileId);
+    }
+
+    public void closeSilently() {
+        try {
+            close();
+        } catch (Exception e) {
+            logger.warn("File close failed for {},{},{}", name, share, fileId, e);
```
hierynomus: Pipe close?

pepijnve: Yep copy/paste mistake

pepijnve:
This PR is very much preliminary work for MSRPC support. I did some quick manual testing against a Windows server, connecting to `IPC$`, waiting and peeking `srvsvc` and trying to write some data to it.
The individual operations work correctly. Not sure if you can actually do anything useful with it yet.

src/main/java/com/hierynomus/msfscc/fsctl/FsCtlPipePeekResponse.java
```diff
@@ -0,0 +1,82 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hierynomus.msfscc.fsctl;
+
+import com.hierynomus.protocol.commons.EnumWithValue;
+import com.hierynomus.protocol.commons.buffer.Buffer;
+
+public class FsCtlPipePeekResponse {
```
hierynomus: Can you add the JavaDoc for the sections where these messages are found in the specs.

pepijnve:
Added references to the relevant sections in MS-FSCC

pepijnve:
@hierynomus any more changes needed for this one?
Could be useful for your DFSC work since it adds some higher level methods to perform IOCTLs.

pepijnve:
Added javadoc and refined the ioctl functions a bit

src/main/java/com/hierynomus/msfscc/fsctl/FsCtlPipeWaitRequest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public long getTimeout() {
+        return timeout;
+    }
+
+    public void write(Buffer buffer) {
+        buffer.putUInt64(timeout / 100L);
+
+        int nameLengthPos = buffer.wpos();
+        buffer.putUInt32(0);
```
hierynomus: This can be pre-calculated without all the jumping in the buffer.

src/main/java/com/hierynomus/smbj/share/PipeShare.java
```diff
@@ -0,0 +1,107 @@
...
+import com.hierynomus.smbj.common.SMBBuffer;
+import com.hierynomus.smbj.common.SmbPath;
+import com.hierynomus.smbj.io.ArrayByteChunkProvider;
+
+import java.util.Set;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+
+public class PipeShare extends Share {
+    private static final int FSCTL_PIPE_WAIT                         = 0x00110018;
```
hierynomus: Spacing?

src/main/java/com/hierynomus/smbj/share/PipeShare.java
```diff
@@ -0,0 +1,107 @@
...
+     * @param timeout     the amount of time to wait until an instance is available
+     * @param timeoutUnit the unit in which timeout is specified
+     * @return true if an instance of the pipe is available; false if a timeout occurred
+     * @throws SMBApiException if an error occurs while waiting for an instance of the pipe to become available
+     */
+    public boolean waitForPipe(String name, long timeout, TimeUnit timeoutUnit) {
+        long timeoutMs = timeoutUnit.toMillis(timeout);
+
+        SMBBuffer buffer = new SMBBuffer();
+        new FsCtlPipeWaitRequest(name, timeoutMs / 100, timeoutMs > 0).write(buffer);
```
hierynomus: `ms / 100` == deciseconds?

src/main/java/com/hierynomus/smbj/share/PipeShare.java
```diff
@@ -0,0 +1,107 @@
...
+
+        Future<SMB2IoctlResponse> responseFuture = ioctlAsync(FSCTL_PIPE_WAIT, true, new ArrayByteChunkProvider(buffer.getCompactData(), 0));
+        SMB2IoctlResponse response;
+        if (timeoutMs > 0) {
+            // Wait a little bit longer than the requested timeout to allow the server to respond with STATUS_IO_TIMEOUT
+            timeoutMs += + 20;
```
hierynomus: `+= + 20`?

src/main/java/com/hierynomus/msfscc/fsctl/FsCtlPipeWaitRequest.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * [MS-FSCC] 2.3.29 FSCTL_PIPE_WAIT Request
+ */
+public class FsCtlPipeWaitRequest {
+    private String name;
+    private long timeout;
+    private boolean timeoutSpecified;
+
+    public FsCtlPipeWaitRequest(String name, long timeout, boolean timeoutSpecified) {
```
hierynomus: Which unit does timeout have? Setting it you divide by 100, writing you divide by 100 another time... Let's use TimeUnit to keep it clear?

src/main/java/com/hierynomus/smbj/share/PipeShare.java
pepijnve: See the comment right above it. 20ms is somewhat arbitrary. Just needs to be a bit longer than the timeout value we're passing to the server.

src/main/java/com/hierynomus/smbj/share/PipeShare.java
pepijnve: I'm not the one that makes this stuff up 😄 See https://msdn.microsoft.com/en-us/library/cc232126.aspx.

> Timeout (8 bytes): A 64-bit signed integer that specifies the maximum amount of time, in units of 100 milliseconds, that the function can wait for an instance of the named pipe to be available.

src/main/java/com/hierynomus/msfscc/fsctl/FsCtlPipeWaitRequest.java
pepijnve: Yep good idea. Used to be done internally, but adding TimeUnit makes sense when the struct is public.

src/main/java/com/hierynomus/smbj/share/PipeShare.java
hierynomus: Let's make it some magic constant, and the `+= +` looked a bit weird ;)

src/main/java/com/hierynomus/msfscc/fsctl/FsCtlPipeWaitRequest.java
pepijnve: Is there a method that I can use to calculate the length? I know it's essentially `string.length() * 2` for UTF-16, but I don't think it's a good idea to duplicate that calculation everywhere.
I wrote the code this way to keep things decoupled. Buffer#putString encodes the string, FsCtlPipeWaitRequest#write doesn't need to care how it's done exactly. It's length value will always be correct.

src/main/java/com/hierynomus/smbj/share/PipeShare.java
hierynomus: Weird guys those romans 😁

src/main/java/com/hierynomus/smbj/share/PipeShare.java
pepijnve: And now the penny drops. Yeah that `+= +` is a mistake. That just looks ridiculous.

src/main/java/com/hierynomus/msfscc/fsctl/FsCtlPipeWaitRequest.java
pepijnve: The double divide by 100 is a bug. Thanks for spotting that.

pepijnve:
Review comments integrated except for the name length one.

src/main/java/com/hierynomus/smbj/share/Share.java
```diff
@@ -42,6 +44,11 @@
 import java.util.concurrent.atomic.AtomicBoolean;
 
 public class Share implements AutoCloseable {
+    private static final SMB2FileId ROOT_ID = new SMB2FileId(
```
pepijnve: I considered that option. I don't think this is used anywhere except for operations that are targeted at the share itself. That's why it's located in Share. Making it a public constant on SMB2FileId seems to communicate that it's generally useful which I don't think it is.
No objections to moving it though if you prefer to have a private constructor. I don't think there's much benefit in making the constructor private though. You can always still create an instance using the static factory method.

pepijnve:
I'll add the string length methods to Buffer.

Writing to a temp buffer doesn't seem like a good idea. That just results in more allocations and memory copies. Plus the write methods would then need to know which endianness to use which is now determined by the input parameter.

pepijnve:
On second thought, I'm still not convinced the string length thing is an improvement. It basically means you have to do the String to byte[] conversion twice; once to calculate the length and once to actually use the bytes. For UTF-16 it's pretty predictable, but for UTF-8 I don't think there's an easy way to calculate the byte length without actually performing the conversion.
Is that really an improvement over encoding and measuring the actual length? Encoding twice is much more expensive computationally than 2 or 3 int field writes.

hierynomus:
LGTM :)

PR #100, opened 2017-06-21
Rename Config to SmbConfig and expose SocketFactory in Config

hierynomus:
The class is renamed because `Config` is very generic and used by a lot of libraries. It prevents needing to fully qualify the classname in case of conflict.

Also this exposes the SocketFactory which allows you to more specifically tweak the Sockets if needed. We still retain the `soTimeout`setting as you can set that without the need to tweak the entire `SocketFactory`, which makes for an overall better experience.


hierynomus:
This fixes #91

PR #101, opened 2017-06-21
Remove all indefinite waits for responses

pepijnve:


pepijnve:
Supersedes #86 which was accidentally closed.

PR #108, opened 2017-06-23
Handle EOF correctly in File#read

### LINKED ISSUE #107: File.read(buffer, fileOffset, offset, length) does not handle EOF status.
File.read(buffer, fileOffset, offset, length) calls share.read(fileId, offset, length) which does not handle EOF, results in a RuntimeException.

The FileInputStream handles EOF in loadBuffer()

pepijnve:
Fixes #107

### DIFF
--- a/src/main/java/com/hierynomus/smbj/share/File.java
+++ b/src/main/java/com/hierynomus/smbj/share/File.java
@@ -15,6 +15,7 @@
  */
 package com.hierynomus.smbj.share;
 
+import com.hierynomus.mserref.NtStatus;
 import com.hierynomus.mssmb2.SMB2FileId;
 import com.hierynomus.mssmb2.messages.SMB2ReadResponse;
 import com.hierynomus.mssmb2.messages.SMB2WriteResponse;
@@ -102,7 +103,7 @@ public class File extends DiskEntry {
      * Read data from this file starting at position fileOffset into the given buffer.
      * @param buffer the buffer to write into
      * @param fileOffset The offset, in bytes, into the file from which the data should be read
-     * @return the actual number of bytes that were read
+     * @return the actual number of bytes that were read; or -1 if the end of the file was reached
      */
     public int read(byte[] buffer, long fileOffset) throws IOException {
         return read(buffer, fileOffset, 0, buffer.length);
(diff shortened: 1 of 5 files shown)


src/main/java/com/hierynomus/smbj/share/File.java
```diff
@@ -114,14 +115,22 @@ public int read(byte[] buffer, long fileOffset) throws IOException {
...
+     * @return the actual number of bytes that were read; or -1 if the end of the file was reached
      */
     public int read(byte[] buffer, long fileOffset, int offset, int length) throws IOException {
         SMB2ReadResponse response = share.read(fileId, fileOffset, length);
-        byte[] data = response.getData();
-        int bytesRead = Math.min(length, data.length);
-        System.arraycopy(data, 0, buffer, offset, bytesRead);
-        return bytesRead;
+        if (response.getHeader().getStatus() == NtStatus.STATUS_END_OF_FILE) {
+            return -1;
```
bli-r7: Would it be ignoring the actual number of bytes that were already read into the buffer before EOF if -1 is returned here?

pepijnve: No, there isn't any data if you get any status code other than STATUS_SUCCESS.

pepijnve: See https://msdn.microsoft.com/en-us/library/cc246729.aspx?f=255&MSPPError=-2147217396 for the details.
SMBJ sets MinimumCount in the read request to 1. So we should only be getting END_OF_FILE if there is no data left. In any case, SMBJ currently doesn't parse the data even if there is any unless STATUS_SUCCESS is returned. See com/hierynomus/mssmb2/messages/SMB2ReadResponse.java:38

hierynomus:
Do you ever rest Pepijn :D?! I will try to take a look at this over the
weekend. Can you add some tests?

Have a look at ckhermann-dfs branch DFSTest to see how you can stub out the
connection.

Op 23 jun. 2017 9:03 p.m. schreef "Pepijn Van Eeckhoudt" <
notifications@github.com>:

pepijnve:
Thanks for merging already. I didn't find time to add some unit tests for this yet...

hierynomus:
It's not merged yet I think, or I accidentally pressed the  button

Op 26 jun. 2017 10:19 a.m. schreef "Pepijn Van Eeckhoudt" <
notifications@github.com>:

Thanks for merging already. I didn't find time to add some unit tests for
this yet...

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/hierynomus/smbj/pull/108#issuecomment-310993794>, or mute
the thread

pepijnve:
Edited comment already. Misread...

pepijnve:
Add some basic file reading tests inspired by the DFS ones.
I made use of the new TransportLayerFactory and injected responses that way instead of stubbing the client and connection.

hierynomus:
Yes, with the decoupling done by @oranoran that is way cleaner indeed!

hierynomus:
I assume you've checked that without the fix at least one of the tests is red?

pepijnve:
Yep, fails with the NPE reported in #107

PR #111, opened 2017-06-29
Async TCP transport

oranoran:


hierynomus:
Please use spock framework / groovy for testing

src/main/java/com/hierynomus/smbj/connection/AsyncDirectTcpTransportFactory.java
```diff
@@ -0,0 +1,76 @@
...
+        this.group = group;
+    }
+    
+    /**
+     * Convenience method for setting up a just single thread that will do all
+     * asynchronous I/O, unlike the default configuration which spins up
+     * multiple I/O processing threads. Using this group makes it easier to
+     * understand which thread does what when debugging.
+     */
+    public void setChannelGroupWithSingleThread(String threadName) {
```
hierynomus: Feels like test code, instead of production code...

src/main/java/com/hierynomus/smbj/lock/CrossThreadLock.java
```diff
@@ -0,0 +1,77 @@
...
+package com.hierynomus.smbj.lock;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * A simple, non-reentrant lock mechanism, which unlike {@link ReentrantLock} allows a thread to release a lock acquired
+ * by a different thread.
+ */
+public class CrossThreadLock {
```
hierynomus: I'm pretty sure this class is non-safe in the JMM. Why is this needed, and aren't there any OOB primitives for this?

src/main/java/com/hierynomus/smbj/transport/tcp/AsyncDirectTcpTransport.java
```diff
@@ -0,0 +1,149 @@
...
+public class AsyncDirectTcpTransport<P extends Packet<P, ?>> implements TransportLayer<P> {
+    public static final int DEFAULT_CONNECT_TIMEOUT = 5000;
+
+    private final Logger logger = LoggerFactory.getLogger(this.getClass());
+    private final PacketHandlers<P> handlers;
+    private final AsynchronousSocketChannel socketChannel;
+    private final AsyncPacketReader<P> packetReader;
+    private int soTimeout = 0;
+
+    // AsynchronousSocketChannel doesn't support concurrent writes
```
hierynomus: And there is a good reason for that probably.... So what is it you're trying to accomplish with the CTL?

src/main/java/com/hierynomus/smbj/connection/AsyncDirectTcpTransportFactory.java
oranoran: Actually it started out as test code, but I moved it into production as a simple deployment option, because I think users may find it useful too for simplifying how they use the library.

src/main/java/com/hierynomus/smbj/lock/CrossThreadLock.java
oranoran: Not sure what you mean about being non-safe. It passed some stress-testing using the test class, with larger parameters than the default.
Didn't find an OOB lock implementation that would work across threads. Actually I was kind of surprised that ReentrantLock doesn't. Maybe there is one in third-party code, I preferred not to introduce new dependencies.

src/main/java/com/hierynomus/smbj/transport/tcp/AsyncDirectTcpTransport.java
oranoran: I think it's a limitation, or rather a design choice on their part to build a simple AsynchronousSocketChannel and leave it to their users to worry about concurrent writes. Their JavaDoc states that concurrent writes will result in exceptions.

A possible use case where this will be a problem in smbj is where multiple sessions do concurrent writes on the same connection.
Possible design options for handling this situation in smbj are:
1. Not to address this and leave it up to the smbj user to synchronize their writes. This would require users to somehow release locks after the future completes, so they would also need to tackle cross-thread lock and release.
2. Introduce a queue of pending writes inside AsyncDirectTcpTransport - this is probably the most performant solution (no locking while waiting on I/O) but introduces a bit more implementation complexity and additional state per connection.
3. Use a lock for pending writes inside AsyncDirectTcpTransport - this is inline with a similar solution already present in DirectTcpTransport.

I think option 1 is less preferred, and don't have a strong opinion between 2 and 3. Went with 3 just based on the pattern already in use in DirectTcpTransport.

oranoran:
About spock framework / groovy for testing - I don't have experience with Spock, but to my (shallow) understanding it has a structured model that makes sense for testing logical processing - feeding a tested class with inputs and examining the outputs. It feels like less of a match for stress-testing, which is what the test is really about.

The test in question is really a stress test for the lock class, which helped uncover some subtle issues as I was writing it. It sets up a bunch of threads and let them compete for locks for a while, then looks at their liveness state and counters.

Any advice how to make this test more compliant?

hierynomus:
For the lock, why not Semaphore? 

oranoran:
Semaphore could work, will try.

src/main/java/com/hierynomus/smbj/connection/AsyncDirectTcpTransportFactory.java
hierynomus: Let's make 2 static methods then, `newAsyncTransportFactory()` and `newSingleThreadedAsyncTransportFactory()`

src/main/java/com/hierynomus/smbj/transport/tcp/AsyncDirectTcpTransport.java
hierynomus: I think you mean this section from the JavaDoc

> Channels of this type are safe for use by multiple concurrent threads. They support concurrent reading and writing, though at most one read operation and one write operation can be outstanding at any time. If a thread initiates a read operation before a previous read operation has completed then a ReadPendingException will be thrown. Similarly, an attempt to initiate a write operation before a previous write has completed will throw a WritePendingException.

The reason why this is the case is that having 2 writes to a socket means that you have no clue how the bytes are interleaved. You must finish 1 write before you can write the next, else you get garbage on the line. This does mean you need to have external synchronization or a queue indeed. But that should be handled on the transport level. I really don't understand your option 1.
Option 2 makes life simple, and could work very well, for instance using a `LinkedBlockingQueue`. I think it isn't more complex, but maybe even simpler
For option 3, I still don't understand why you need a 'CrossThreadLock'. What is the reasoning for not using a regular ReentrantLock?

pepijnve: Do I understand it correctly that this implementation returns early before the write has completed? If so, then the behaviour of the regular synchronous transport and async one are very different and you can't use one as a drop-in replacement for the other.

@hierynomus The documentation of `TransportLayer#write(P packet)` doesn't state it's post-conditions explicitly, which it probably should. Upon return, should the packet have been written out to the network? If not, how do you get notification of this? If that guarantee should be given, then `AsyncDirectTcpTransport#write` might as well be made synchronized and block until the write is complete.

src/main/java/com/hierynomus/smbj/connection/AsyncDirectTcpTransportFactory.java
oranoran: Isn't having factory methods for the factory object a bit too many layers?

src/main/java/com/hierynomus/smbj/transport/tcp/AsyncDirectTcpTransport.java
oranoran: @pepijnve Many elements in the existing transport and SMB layer design are already asynchronous. The new suggested mode is really about shifting this asynchronous design from a thread-per-connection model to a shared thread model.

On the transport layer, the "read" side of the existing transport is already implemented asynchronously, using a dedicated thread per connection. On the SMB layer, every operation returns a Future object, which is really an async design.

The only difference in functionality with the new transport, as you point out, is that now writes return early. I think the only impact of this difference is what happens when a write fails. Since there is already a facility in place for reporting asynchronous errors, currently named PacketReader, the async implementation simply reuses it also for reporting async write errors.
This raises a valid concern about naming and interface responsibilities, since write errors get reported to a PacketReader. I think maybe it's time to refactor PacketReader into two separate interfaces - PacketReader and a TransportErrorHandler. The TransportErrorHandler part would be used for reporting any asynchronous transport error, for both reading and writing. Does that make sense?

Is there any other use case where you expect things to break by by changing the write model?

src/main/java/com/hierynomus/smbj/transport/tcp/AsyncPacketReader.java
```diff
@@ -0,0 +1,190 @@
...
+
+    private static final int READ_BUFFER_CAPACITY = 9000; // Size of a Jumbo frame
+    private static final int HEADER_SIZE = 4;
+
+    private final ByteBuffer readBuffer;
+    private int bytesWaitingToBeProcessed = 0;
+
+    private static final int NO_PACKET_LENGTH = -1;
+
+    private int currentPacketLength = NO_PACKET_LENGTH;
```
hierynomus: I don't like these packet instance specific fields, in the Reader. Feels like you're mixing stuff up here.

oranoran: The question is what happens if an incoming SMB packet is split into more than one read chunk. In theory at least, I don't think there is a guarantee that this won't happen. In such a case, without maintaining state between reads, the packet will fail to be processed.

hierynomus: Indeed, that is possible. However I think you should encapsulate this logic instead of flattening it out in the PacketReader.

oranoran: The AsyncPacketReader class is very specific for this purpose exactly. All it does is make read() calls and track the read state, calling handler.handle() when a new packet has been read. It's hard for me to imagine how part of this functionality could be meaningfully split out. Can you offer a patch or something?

hierynomus:
Here is a proposed change to the PR.

Note however that I think your code is suffering from a potential StackOverflowError, as you're recursively calling the `readPacket` method... It won't go very fast, but at first glance it's definitely there. The patch does not address that, it just illustrates what I was hinting at.
[use_attachment.txt](https://github.com/hierynomus/smbj/files/1122355/use_attachment.txt)

oranoran:
Thanks @hierynomus. I will look at the patch later, but wanted to comment quickly on the concern about StackOverflowError.
I don't think there is a bug there. Please note that calls to readPacket are done from the CompletionHandler. Each invocation of the CompletionHandler is a separate task in the channel's executor, running independently of the previous invocation that triggered it. The completion handler may or may not reuse the same actual thread, depending on the thread pool being used, but the stack will start fresh.

hierynomus:
Thanks @oranoran, you're right, missed that :)
Have a look at the patch and let's try to get this finished.

hierynomus:
Thanks for the PR and the improvements! Merging...

Issue #94, opened 2017-06-20
DiskShare.list() not working if client requires signing


Pernicies:
If the method list of the DiskShare class is used with a SMBClient instance, that was configured to require signing, an exception will be thrown as an invalid packet signature was received.

Tracking it down in the debugger the exception is raised in the queryDirectory method of the Directory class when the server responds with a message having the nt-status STATUS_NO_MORE_FILES.

This behaviour is/was reproducable on shares hosted on Windows 10 and Windows Server 2012 machines.

hierynomus:
I've seen this happen before on non- NtStatus.SUCCESS responses.

Will dig into it.

2017-06-20 12:57 GMT+02:00 Pernicies <notifications@github.com>:

bli-r7:
This is likely related to https://github.com/hierynomus/smbj/issues/112

PR #113, opened 2017-06-30
Correctly read error responses (Fixes #112 #94)

### LINKED ISSUE #112: Non-successful response would fail signature verification.
The signature is not generated with the entire message if the response status is not STATUS_SUCCESS for at least a few message type (SMB2ReadResposne and SMB2QueryDirectoryResponse)

The packet.getMessageEndPos()'s return value in PacketSignatory#verify is off.

This issue can be reproduced by requiring signature in the SmbConfig and read from an empty file.

hierynomus:


hierynomus:
@pepijnve @bli-r7 @Pernicies Could you verify whether this fixes signature validation. My initial tests show it should, but would be great to have some additional checks.

Pernicies:
I can confirm that the signature validation is now working as expected in my application.

pepijnve:
Change itself looks good. I'm not at work for two weeks. I'll run some tests against our NetApp servers when I get back. Those were causing signature verification errors during testing.

bli-r7:
The fix worked for me as well

PR #126, opened 2017-07-14
Added helper methods for writing files to share

mtamassia:
Added 3 helper methods to DiskShare to facilitate uploads:
 public int write(String path, InputStream is, boolean overwrite)
 public int write(String path, java.io.File localFile, boolean overwrite)
 public int write(String path, java.io.File localFile)

### DIFF
--- a/src/main/java/com/hierynomus/smbj/share/DiskShare.java
+++ b/src/main/java/com/hierynomus/smbj/share/DiskShare.java
@@ -36,8 +36,10 @@ import com.hierynomus.smbj.common.SMBApiException;
 import com.hierynomus.smbj.common.SMBBuffer;
 import com.hierynomus.smbj.common.SMBRuntimeException;
 import com.hierynomus.smbj.common.SmbPath;
+import com.hierynomus.smbj.io.InputStreamByteChunkProvider;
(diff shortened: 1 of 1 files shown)


mtamassia:
Apologies for all the Line Break changes triggering the diff to flag all lines. I only added 53 lines of code with the 3 functions...

hierynomus:
Hi @mtamassia,

As it is, I will not accept the PR in this state. This messes up the entire file.

Furthermore, I do not agree that these functions belong in the `DiskShare`. I would be ok-ish with Adding a `SmbFiles.copy(File source, DiskShare dest, String destPath)` or something like that.

mtamassia:
No problem @hierynomus. 

I will change git line endings behavior on my end as well so the future PRs are clean.

I can add a static SmbFiles.copy method. Would you want that into its own SmbFiles class inside main.java.com.hierynomus.smbj, inside main.java.com.hierynomus.smbj.share or would you want somewhere else?

hierynomus:
First idea would be com.hierynomus.smbj.utils

Op 14 jul. 2017 22:03 schreef "mtamassia" <notifications@github.com>:

mtamassia:
Closing this PR as I sent you a new one.

PR #140, opened 2017-08-01
Ensure Connection#close and Session#close release the relevant resources

bli-r7:
Move the transport.disconnect and event publication to finally block in Connection#close and Session#close.


bli-r7:
@hierynomus Could you review this PR when you have time and let me know what you think? It is particularly important for us that the connection can be reliably closed.

bli-r7:
@hierynomus Would you please merge this PR if there's no outstanding issues/comments when you have time?

Thanks.

PR #143, opened 2017-08-03
Recover from STATUS_BUFFER_OVERFLOW on named pipe transactions.

pmiseiko-r7:
Add a read method to the named pipe class that returns the SMBv2 read response object.  Update the SMBv2 read response object to parse and return the remaining data to be read from the named pipe.

When a transaction call cannot complete because its buffer is not large enough, it is possible to write the call to the named pipe and than read it back in chunks.  Some versions of Windows (this was experienced on Windows 2012 R2) will return INVALID_PARAMETERS if the transaction buffer is too large but are otherwise okay with supporting large read buffers.  If the transaction buffer is too small and BUFFER_OVERFLOW is returned, you have to drain the pipe with SMBv2 read calls otherwise the next transaction will break because the named pipe buffer on the server still contains the data remaining from the previous transaction call.

Example code:
```
final int callID = this.callID++;
final byte[] requestBytes = request.marshal(callID);
try {
    final byte[] responseBytes = new byte[MAXIMUM_TRANSACT_SIZE];
    namedPipe.transact(requestBytes, responseBytes);
    return request.unmarshal(responseBytes, callID);
} catch (final SMBApiException exception) {
    final NtStatus status = exception.getStatus();
    if (status.equals(NtStatus.STATUS_BUFFER_OVERFLOW)) {
        for (;;) {
            final SMB2ReadResponse readResponse = namedPipe.read(MAXIMUM_READ_SIZE);
            if (0 >= readResponse.getDataRemaining()) {
                break;
            }
        }
    }
    final ByteArrayOutputStream responseBytes = new ByteArrayOutputStream();
    namedPipe.write(requestBytes);
    for (;;) {
        final SMB2ReadResponse readResponse = namedPipe.read(MAXIMUM_READ_SIZE);
        responseBytes.write(readResponse.getData());
        if (0 >= readResponse.getDataRemaining()) {
            break;
        }
    }
    return request.unmarshal(responseBytes.toByteArray(), callID);
}
```


hierynomus:
Instead of exposing the inner messages of the SMB2 protocol, isn't htere a clean way to handle this internally as more of an API?

bli-r7:
@hierynomus Could you review this PR when you have time as well? We observed issues with named pipe on certain target where we need to handle STATUS_BUFFER_OVERFLOW this way.

The target we had issue with is a Windows Server 2012 in this particular case though there are Windows Server 2012 we tested that works.

hierynomus:
@bli-r7 I commented earlier this afternoon that I don't like exposing the Responses. I'd rather handle this internally

bli-r7:
@hierynomus Ah sorry, I didnt' refresh the page before I commented.

hierynomus:
If that is a standard and supported way of handling this, it makes sense to
add that mechanism in the library.

Op 9 aug. 2017 6:14 p.m. schreef "pmiseiko-r7" <notifications@github.com>:

pmiseiko-r7:
I don't know if it's a standard way to handle it, but it works. :)

I'll look at a PR later this week with that enhancement and we can talk over the differences.

pepijnve:
@hierynomus the documentation of [TransactNamedPipe](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365790%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396) hints at this approach as well.

> If the message to be read is longer than the buffer specified by the nOutBufferSize parameter, TransactNamedPipe returns FALSE and the GetLastError function returns ERROR_MORE_DATA. The remainder of the message can be read by a subsequent call to ReadFile, ReadFileEx, or PeekNamedPipe.

When using the transact overloads that accept an output buffer there's not much this method can do. `transact(byte[] inBuffer)` could implement this logic automatically. The other overloads might have to return some more information than 'number of bytes written'.

hierynomus:
@pmiseiko-r7 Any idea when you're able to update this PR? It seems that your colleague @bli-r7 is waiting for it ;)

pmiseiko-r7:
@hierynomus created a new pull request for this (https://github.com/hierynomus/smbj/pull/155).

PR #144, opened 2017-08-03
SMBClient should close connection if fails to connect.

bli-r7:
The caller of SMBClient doesn't have the reference to Connection to close it so it needs to be done in SMBClient.


hierynomus:
You can use the protocol.commons.IOUtils.closeQuietly instead of the inner try-catch block. 

hierynomus:
Implemented this by using `closeSilently` in 49563d1

Issue #138, opened 2017-08-01
Remove the mbassador library log message: "No error handler has been configured to handle exceptions during publication."


SuperPat45:
Hi,

The mbassador library add this log this message (since version 1.1.7) each time a new SMBClient is instanciated:
> INFO: No error handler has been configured to handle exceptions during publication.
> Publication error handlers can be added by IBusConfiguration.addPublicationErrorHandler()
> Falling back to console logger.

I found this message can be easily removed by changing the SMBEventBus constructor:
```
public class SMBEventBus {

    private PubSubSupport<SMBEvent> wrappedBus;

    public SMBEventBus() {
        this(new SyncMessageBus<SMBEvent>());
    }
//....
```
by:
```
public class SMBEventBus {

    private PubSubSupport<SMBEvent> wrappedBus;

    public SMBEventBus() {
        this(new SyncMessageBus<SMBEvent>(new IPublicationErrorHandler.ConsoleLogger()));
    }
//....
```

mtamassia:
@SuperPat45 I submitted a Pull Request #146 with the changes you have above. Thank you.

SuperPat45:
Thanks

PR #146, opened 2017-08-04
Issue #138 Remove the mbassador library warning messages from log

mtamassia:
Constructor changed to pass Error Handler, expected since mbassador  version 1.1.7


hierynomus:
Closing this, because I think it is better to log the error to SLF4J instead of relying on System.out/err. This is fixed in commit f2b17cd

PR #149, opened 2017-08-08
Implementation for file setLength (truncate file)

yin19941005:
Implementing the setLength function for file. When writing a large file, setLength before actual writing the file would improve the performance. Also can be used for truncating the files.

### DIFF
--- a/src/main/java/com/hierynomus/smbj/share/File.java
+++ b/src/main/java/com/hierynomus/smbj/share/File.java
@@ -156,8 +156,12 @@ public class File extends DiskEntry {
      * @throws SMBApiException
      */
     public void setLength(long endOfFile) throws SMBApiException {
-        FileEndOfFileInformation endOfFileInfo = new FileEndOfFileInformation(endOfFile);
-        this.setFileInformation(endOfFileInfo);
+        try {
+            FileEndOfFileInformation endOfFileInfo = new FileEndOfFileInformation(endOfFile);
+            this.setFileInformation(endOfFileInfo);
+        } catch (IllegalArgumentException iae) {
+            logger.warn("setLength : " + iae);
+        }
     }
 
     public InputStream getInputStream() {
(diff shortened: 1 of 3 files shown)


src/main/java/com/hierynomus/smbj/share/File.java
```diff
@@ -148,6 +150,16 @@ public void read(OutputStream destStream, ProgressListener progressListener) thr
         is.close();
     }
 
+    /***
+     * The function for truncate or set file length for a file
+     * @param endOfFile 64-bit signed integer in bytes, MUST be greater than or equal to 0
+     * @throws SMBApiException
+     */
+    public void setLength(long endOfFile) throws SMBApiException {
```
hierynomus: Add the boundary check to the method (ie. `endOfFile > 0`)

yin19941005: This is my fault for didn't handle the checking. However, I got a slightly different point of view. As [MS-FSCC document](https://msdn.microsoft.com/en-us/library/cc232067.aspx), all the FileEndOfFileInformation information class should handle the checking for the EndOfFile. So, I think it is better to implement the checking in the information class instead of the method. Do you agree with this?

hierynomus: Agreed.

src/main/java/com/hierynomus/smbj/share/File.java
```diff
@@ -148,6 +150,20 @@ public void read(OutputStream destStream, ProgressListener progressListener) thr
...
+        try {
+            FileEndOfFileInformation endOfFileInfo = new FileEndOfFileInformation(endOfFile);
+            this.setFileInformation(endOfFileInfo);
+        } catch (IllegalArgumentException iae) {
```
hierynomus: Don't do this. That's pretty bad form for a library.

yin19941005: Sorry, I didn't got what you mean clearly. Do you mean I shouldn't catch the exception and just let it throws the exception to the caller (let the caller to handle the exception by itself)?

hierynomus: Yes, indeed.

yin19941005: Ok, I got it now. Indeed is a bad form for library.

PR #155, opened 2017-08-14
Update the IOCTL logic to handle a buffer overflow status.

pmiseiko-r7:
Update the IOCTL logic to handle a buffer overflow status.  STATUS_BUFFER_OVERFLOW is now considered a success status.  If the IOCTL operation returns a STATUS_BUFFER_OVERFLOW, the named pipe is READ until no more data is remaining.  This change attempts to store the complete output of an IOCTL operation in the provided output buffer.  If the provided output buffer is too small, the named pipe is READ until no more data is remaining, and the original IOCTL exception is thrown to the caller.  This informs the caller that a buffer overflow occurred, preserving legacy behavior, however with this change the buffer overflow is a local issue instead of a remote issue.  The named pipe must be READ until no more data is remaining otherwise the output buffer of a future IOCTL call will contain data from the previous IOCTL call.

Note 1:
Have not tested against all versions of Windows yet (only tested against Windows 8.1).

Note 2:
This is an evolution of PR https://github.com/hierynomus/smbj/pull/143.


pmiseiko-r7:
One unfortunate thing of this change is the other ioctl method is not fixed:
```
byte[] ioctl(SMB2FileId fileId, int ctlCode, boolean isFsCtl, byte[] inData, int inOffset, int inLength, int maxOutputResponse)
```

This logic also seems like a great candidate for a dynamic byte buffer instead of passing in a large buffer with every call.  Do you have an existing dynamic byte buffer you're using?

src/main/java/com/hierynomus/smbj/share/Share.java
```diff
@@ -260,12 +262,55 @@ public int ioctl(int ctlCode, boolean isFsCtl, byte[] inData, int inOffset, int
     }
 
     int ioctl(SMB2FileId fileId, int ctlCode, boolean isFsCtl, byte[] inData, int inOffset, int inLength, byte[] outData, int outOffset, int outLength) {
```
hierynomus: Wrong place for this logic. The BUFFER_OVERFLOW logic is only for NamedPipe, not for all other ioctl ops. Move this into `NamedPipe.ioctl`

pepijnve:
I was under the impression this should only ve done for pipe transactions even, not for all ioctls on pipes.

hierynomus:
I think you're right indeed. NamedPipe.transact is the right place

Op 15 aug. 2017 8:54 a.m. schreef "Pepijn Van Eeckhoudt" <
notifications@github.com>:

I was under the impression this should only ve done for pipe transactions
even, not for all ioctls on pipes.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/hierynomus/smbj/pull/155#issuecomment-322391170>, or mute
the thread

pmiseiko-r7:
Happy to move it.  No issues with the implementation?

hierynomus:
Haven't looked in depth yet, but you need to refactor a bit here and there if you move it ;)

pmiseiko-r7:
@hierynomus moved the change into NamedPipe.java; looks a lot better there.  The transact method does not have a bounded read buffer which could be a security concern for the client if the server wants to be malicious; there's a few other pieces of the code that could be cleaned up, but want you feedback before moving forward with any changes.  Thoughts?

src/main/java/com/hierynomus/smbj/share/NamedPipe.java
```diff
@@ -111,7 +120,48 @@ public int read(byte[] buffer, int offset, int length) {
...
+
+                final byte[] readData = readResponse.getData();
+                try {
+                    outBuffer.write(readData);
+                } catch (final IOException exception) {
+                    throw new SMBRuntimeException(exception);
+                }
+
+                final NtStatus readStatus = readResponse.getHeader().getStatus();
+                if (!readStatus.equals(NtStatus.STATUS_BUFFER_OVERFLOW)) {
```
hierynomus: Is this correct? Shouldn't it stop on EOF, not on something different than BUFFER_OVERFLOW?

src/main/java/com/hierynomus/smbj/share/NamedPipe.java
```diff
@@ -111,7 +120,48 @@ public int read(byte[] buffer, int offset, int length) {
...
+        final byte[] ioctlOutputBuffer = ioctlResponse.getOutputBuffer();
+        try {
+            outBuffer.write(ioctlOutputBuffer);
+        } catch (final IOException exception) {
+            throw new SMBRuntimeException(exception);
+        }
+
+        final NtStatus ioctlStatus = ioctlResponse.getHeader().getStatus();
+        if (ioctlStatus.equals(NtStatus.STATUS_BUFFER_OVERFLOW)) {
+            for (;;) {
```
hierynomus: Extract to a method (`emptyPipe`?)

src/main/java/com/hierynomus/smbj/share/NamedPipe.java
```diff
@@ -125,14 +175,19 @@ public int read(byte[] buffer, int offset, int length) {
      * @return the number of bytes written to <code>outBuffer</code>
      */
     public int transact(byte[] inBuffer, byte[] outBuffer) {
-        return transact(inBuffer, 0, inBuffer.length, outBuffer, 0, outBuffer.length);
```
hierynomus: Why aren't we calling the method below anymore? Code is exactly the same except for the range-check which is a no-op when called from this method.

src/main/java/com/hierynomus/smbj/share/NamedPipe.java
```diff
@@ -141,7 +196,15 @@ public int transact(byte[] inBuffer, byte[] outBuffer) {
      * @return the number of bytes written to <code>outBuffer</code>
      */
     public int transact(byte[] inBuffer, int inOffset, int inLength, byte[] outBuffer, int outOffset, int outLength) {
-        return ioctl(FSCTL_PIPE_TRANSCEIVE, true, inBuffer, inOffset, inLength, outBuffer, outOffset, outLength);
+        if (outBuffer.length - outOffset < outLength) {
+            throw new IllegalArgumentException();
+        }
+        final byte[] transactOutBuffer = transact(inBuffer);
+        if (transactOutBuffer.length > outLength) {
+            throw new BufferOverflowException();
```
pepijnve: Wouldn't it be more useful to read as much as you can into the provided buffer and in case of buffer overflow return the number of bytes written as a negative value? (or something similar to that; a special return value.) It is then up to the caller to handle this case correctly in the same way the higher level `transact(byte[])` does.

With this implementation the data has already been read and is not recoverable by the caller.

hierynomus: You can retry the `transact` with the correct buffer size, right?

I think throwing an exception is the right thing. If you have no clue of the sizes needed, you can use the higher-level `transact(byte[])`, if you pass in a too small outBuffer, it makes sense to throw the exception and leave the Pipe in a functional state for the next request.

hierynomus: Also as I noted in the PR, this `transact(byte[], byte[])` should just call this method, so we can handle it in one place.

pepijnve: My thinking was that it might make sense to try and mimic [TransactNamedPipe](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365790%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396). That also says in case of overflow that you can get the rest of the data by reading.
Retrying will only work properly if the operation in question is idempotent. Any ioctl that has a side-effect can not necessarily just be attempted again.

hierynomus: Maybe we should then do it differently even and have a lock on the NamedPipe that prevents starting another `transact` until the pipe has been emptied?

pepijnve: Just getting the single user use case working correctly is probably ok. Imagine you do a 'create user' RPC call. You try once, get buffer overflow. Try again and then get user already exists. The operation was performed, we just threw away the answer. Would be kind of weird IMO.

pepijnve: Synchronize a remote object locally doesn't really help much is what I'm trying to say. You can open the same pipe twice locally. Multiple people could have it open. A local mutex on the NamedPipe object doesn't fix this.

hierynomus: Ok, so how is this case then handled in other places? What happens if somebody else leaves stuff in the pipe? What if two commands are executed on the same pipe? How is the interleaving handled?

pepijnve: ¯\_(ツ)_/¯
Looking at RPC calls specifically I would guess that it behaves like a socket on the server side. Every connection has its own input/output stream. How the RPC backend deals with this is implementation dependent I guess. For instance if multiple people connect to `srvsvc` and issues a [NetShareSetInfo](https://msdn.microsoft.com/en-us/library/windows/desktop/bb525389(v=vs.85).aspx) at the same time, it's dependent on the backing SMB server how it synchronises this.

src/main/java/com/hierynomus/smbj/share/NamedPipe.java
pmiseiko-r7: EOF looks to be SUCCESS for a transaction.  Haven't seen a EOF status yet in current testing (Windows Vista, Windows 2008 R2, Windows 8.1, and Windows 10).

pmiseiko-r7: Also this will stop if the response is either EOF or SUCCESS which looks to be correct.  While I haven't seen a EOF, if we ever got one, than that's a good indicator that this code should stop as well.  :)

src/main/java/com/hierynomus/smbj/share/NamedPipe.java
pmiseiko-r7: I don't think it's necessary to move this to another method.  This code is reading the pipe after the transaction failed; so the first thing it does is add the transaction response buffer to the out buffer which is the first part of the named pipe response.  It follows that up with a loop that continues to read until no more data is pending from the named pipe.  In this scenario, all the data is consumed, and we make the least number of calls possible (we don't remake the transaction call) so this implementation is the most efficient solution.  However, because the out buffer is not bounded, a malicious server could in theory cause the client to run out of memory.

I can move this to a separate method, perhaps "readFully", if you want.  Might even make sense in the share itself to be re-used for other situations where the data needs to be fully read.

src/main/java/com/hierynomus/smbj/share/NamedPipe.java
pmiseiko-r7: Will push a fix to revert this line.

src/main/java/com/hierynomus/smbj/share/NamedPipe.java
pmiseiko-r7: I see a great conversation happened here but I'm not sure what the conclusion of that conversation was.  Is a change desired here?  Throw the exception, as is; or return -1?  Note that this particular set of code is not the most efficient since it means the transaction would have to be redone so the caller has to deal with further latency issues.

Another idea here is we could return a negative number that represents the amount of data missing, so for example, if you provided a 64k buffer but 70k of data was read, -4k could be returned to inform the caller an additional 4k is needed to store the data.

This call seems irrelevant to me because we are not using the length to manage the other call above i.e. it will read everything, and indeed it has to or the named pipe becomes desynchronized with the client; but it also stores everything, so as a library this method is doing no favors to the caller.

Perhaps I should break the other method out and give it a parameter that you specify the maximum amount of data you are interested in, and -1 could mean unlimited.  This solves the potential security vulnerability here, while making these other methods useful.

I don't know if you would like to see such changes and/or if you want to seem them for this PR or a future PR.  Let me know.

hierynomus: Another option:
For every `transact` call return a `NamedPipeTransaction` that contains the last status and a method to fetch a new chunk of the data that did not yet fit in the outBuffer.

How does that sound? No _magic_ returned numbers that a user has to interpret

pepijnve: That's indeed the cleaner OO/encapsulated solution to the problem. Special return values are so C99 😉

How far do you go in making this 'idiot-proof'? If someone for instance does

```
NamedPipeTransaction a = pipe.transact();
// a has more data
NamedPipeTransaction b = pipe.transact();
b.readMoreData();
```

How do you deal with that case? Should the second pipe.transact fails because there's still an open transaction? Should it silently flush the pipe before writing the new request? ...

I think I'm trying to say that while this is a stylistic improvement it doesn't really prevent you from shooting yourself in the foot.

hierynomus: I think there you could easily prevent that by having the Pipe maintain the knowledge that a transaction is still ongoing.

As you already said it mimics a socket, so any single connection (instance of NamedPipe) is independent of another. At least if I understood your previous comment correctly the following is allowed (and independent):

```
NamedPipe pipe1 = new SMBClient().connect(...).authenticate(...).connectTree(pipe);
NamedPipe pipe2 = new SMBClient().connect(...).authenticate(...).connectTree(pipe);
pipe1.transact(..);
pipe2.transact(..);
```

Right?

pepijnve: That's my understanding of it at least. Not 100% sure. If not then I would expect this to be dealt using sharing modes just like with files.

The pipe could indeed keep track of any open transactions, but that complexity to the code again. I was questioning whether that's worth the extra effort and complexity or not.

hierynomus: I don't think it is a bad form of complexity as you get a clearer API :). Most devs are used to dealing with and handling transactions (JDBC), so providing a similar mechanism here, which would only allow you to start a single transaction on a pipe which needs to be finished first before starting a new one is not a bad thing I think.

pepijnve: Should the NamedPipeTransaction be a Closeable resource? Then you could use try-with-resources to write something like this

```
try (NamedPipeTransaction t = pipe.transact(...)) {
  // read more data if there is any
}
// NamedPipeTransaction#close flushes the pipe if needed.
```

hierynomus:
Nice and clean, I like!

Op 18 aug. 2017 4:55 p.m. schreef "Pepijn Van Eeckhoudt" <
notifications@github.com>:

src/main/java/com/hierynomus/smbj/share/NamedPipe.java
pmiseiko-r7: We might be over thinking this a bit.  I can't think of a reason a caller would not want the complete buffer returned to them; and since this abstraction wraps the API call to simplify the behavior and reduce code duplication, it seems strange that we would give the caller a situation where they can make a decision about whether or not the entire data is of interest to them.

However, continuing this conversation, return a stream might make more sense here instead of making a new class.  Fundamentally, we want to read the data back from the transaction, and that is what a stream is for.  The benefit of the stream is that it does not need to read all the data up front and instead the data can be read as needed by the caller.  This solves the problem using an existing Java idiom, and downstream consumers can choose to operate on the data as it is read instead of buffering it all for higher efficiency.  This also solves the potential vulnerability discussed earlier, or at least passes it onto the caller to decide how to handle excessive responses.

Lastly, the stream can be closed to flush the pipe.

The biggest issue we have here once we returned a "delayed" read from the pipe is what you two have touched on around synchronization.  Nobody else can transact against the pipe until the previous transaction is complete.  We can solve that for that caller by adding synchronization around the main transact method in this PR.  If we return a stream, than synchronization in the library does not solve it, and the caller must now perform synchronization of their own to prevent multiple transactions against the same pipe.

We shouldn't be trying to guess, or store state, since that can easily lead to defects or deadlocks.  When in doubt, assume the caller will make a mistake, because they most likely will.

How about we have two methods here.  One that returns all the data, i.e. what we have above, that includes synchronization.  Call it the "easy method".  And a second method that returns a stream, use at your own risk, for high efficiency applications.  The other two methods can be removed at that point.

Thoughts?

pmiseiko-r7: Example:
```
    public InputStream _transact(final byte[] inBuffer) {
        return new InputStream() {
            boolean doIOCTL = true;
            ByteArrayInputStream buffer;
            NtStatus status;

            @Override
            public int read() throws IOException {
                if (doIOCTL) {
                    final SMB2IoctlResponse ioctlResponse = share.ioctl(
                        fileId,
                        FSCTL_PIPE_TRANSCEIVE,
                        true,
                        new ArrayByteChunkProvider(inBuffer, 0, inBuffer.length, 0),
                        -1,
                        SUCCESS_OR_BUFFER_OVERFLOW
                    );
                    buffer = new ByteArrayInputStream(ioctlResponse.getOutputBuffer());
                    doIOCTL = false;
                }

                int result = buffer.read();
                if (result == -1 && status.equals(NtStatus.STATUS_BUFFER_OVERFLOW)) {
                    final SMB2ReadResponse readResponse = share.read(
                            fileId,
                            0,
                            share.getReadBufferSize(),
                            SUCCESS_OR_EOF_OR_BUFFER_OVERFLOW
                        );
                    buffer = new ByteArrayInputStream(readResponse.getData());
                    status = readResponse.getHeader().getStatus();
                    result = buffer.read();
                }

                return result;
            }
        };
    }
```

hierynomus: I agree with the "when in doubt, assume they will make a mistake". But that is why I would prefer a `Transaction` like approach. Transactions are like streams a widely understood concept. And having an unclosed stream would actually lead to garbage on the next call, which might be even more undesirable. Also the method is called `transact` so returning a Transaction fits the domain also.

pmiseiko-r7: Returning the stream also fits without the need to create a new class.  Do you like the example above (you didn't comment on it)?  Also the proposal was two methods: easy-mode, does it all for you, no risk to the caller, you get everything in one nice byte buffer, and hard-mode, caller gets a stream (or other class such as the proposed Transaction), callers job to read, callers job to close; closing will flush the buffer (didn't put that in the example above, but it can of course be added).

pmiseiko-r7: The recent change is slightly different than what was discussed here.  All of the recent work hid the transact API from the caller, which meant a caller could not create their own complete transaction implementation and/or could not guarantee that this code would not call a read method if the caller required that guarantee.  The latest change reverted modifications made to the transact methods and introduced a transactFully method.  The transactFully method returns an InputStream that handles the full read and implements a close method that flushes the remainder of the transaction.

hierynomus: I'm still doubting...
1. I think we really need to guard against unwanted garbage in the pipe, ie. ensure that we cannot start a new `transact` without the previous one having been drained/closed. This is easily achievable by adding an `AtomicBoolean transactInProgress` which we can check and throw an `TransactInProgressException` if there is still a transaction ongoing.
2. `close` should definitely flush the `NamedPipe` to that it is ready for a new `transact`
3. We also have a `peek` method. If we'd add a `NamedPipeTransaction` this could move there, as it is really a part of that API. Adding it to the `InputStream` makes less sense (as that has a predefined contract). 

I would suggest the following:

1. `byte[] transact(byte[])`: Handles the buffer overflow automatically, returning you a potentially big `byte[]`
2. `int transact(byte[], byte[])` and `int transact(byte[], int, int, byte[], int, int)`: If a buffer overflow occurs, fill the output array and throw an exception that the client needs to handle, by for instance doing subsequent reads. Any further `transact` calls are blocked until the NamedPipe is emptied (add `drain` method)
3. `InputStream transactStream(byte[])`: Does what you have now, and on `close` drains the `NamedPipe`
4. `NamedPipeTransaction transact?(byte[])`: Returns an object which is slightly richer than an InputStream and allows you to do `peek` operations on the transaction. And `drain` the `NamedPipe` on a `close` (like the `InputStream`)

pmiseiko-r7: \1 We can guard against garbage with the synchronize keyword on the transactFully method.  We need this anyway since it's possible for a multi threaded application to use the named pipe and unintentionally run into an in-progress transact.  They either wrap it with synchronization, or we do it here.
\2 Agree, the input stream does that in the current implementation.
\3 Why do you want to peek the stream?

Your suggestions would hide direct access to the API.  It also creates a state object (NamedPipeTransaction) when the NamedPipe object could be that state since it can not support parallel transactions.

pmiseiko-r7: Synchronization has been added to transactFully.

Do we know if it's illegal to perform multiple transactions before the previous transaction is read from?  I.e. one of the proposals above was to throw TransactInProgressException but is that a limitation of the protocol of this implementation of the protocol?  If a caller knows how to handle the combined output of multiple transactions, than throwing that exception would restrict the caller from doing something that is perfectly valid.  Going to test this.

pmiseiko-r7: Tested against Windows 2008.

Test scenarios:
* namedPipe.write() -> namedPipe.transact() = STATUS_PIPE_BUSY
* namedPipe.write() -> namedPipe.read(zero or partial output buffer) -> namedPipe.transact() = remaining output data from namedPipe.write() plus new data from namedPipe.transact()
* namedPipe.transact(zero or partial output buffer) -> namedPipe.write() -> namedPipe.read() = remaining output data from namedPipe.transact() plus new data from namedPipe.write()

Summary:
If you write() to a NamedPipe you must read() from it at least zero bytes otherwise STATUS_PIPE_BUSY is returned.  If you transact() to a NamedPipe you must read the complete transaction otherwise STATUS_PIPE_BUSY is returned.

The server solves the synchronization problem for us in this case with the exception that you can trick it if you write() and don't read() the entire response.  However, it looks to be perfectly valid to write() multiple operations before calling a single read(); so the synchronization is around transact() but not read() and write().

Based on these tests, the proposals would not protect against synchronization issues.  To fully protect the caller from causing issues with the transact() call, both read() and write() would have to be protected/private methods; but the caller would no longer be able to fully use the NamedPipe API in that case (calling multiple writes before reads).

pmiseiko-r7:
@hierynomus I'm going to close this PR.  We realized this was not the problem we were having and had to re-implement our solution.  The problem we were having was related to fragmentation at the RPC layer (the data traversing the named pipe).  We forked the NamedPipe.java class to implement that solution in smbj-rpc.

You can see the implementation here:
https://github.com/rapid7/smbj-rpc/blob/master/src/main/java/com/rapid7/helper/smbj/share/NamedPipe.java

You can see the usage here:
https://github.com/rapid7/smbj-rpc/blob/master/src/main/java/com/rapid7/client/dcerpc/transport/SMBTransport.java

PR #159, opened 2017-08-16
Provide session setup flags negotiated in SMB2SessionSetup to the Session

tosmun-r7:
It does require a dependency on `SMB2SessionSetup.SMB2SessionFlags`, but allows us to get more information about the `Session`, such as whether or not it is a guest.


hierynomus:
I've added isGuest and isAnonymous, that is according to the spec the only information in the Flags that had not been exposed.

tosmun-r7:
Perfect! Thanks @hierynomus

PR #165, opened 2017-08-18
Issue 163 - Throwing exception when search pattern does not match any file

mtamassia:


mtamassia:
@hierynomus please let me know if you have any questions on this PR

PR #186, opened 2017-09-08
Adding the ability to capture metadata for SMBSession

skhare-r7:
We can grab useful information for an SMB session from the connection setup response and the negotiate protocol response. This can be exposed and made useful for fingerprinting purposes.


hierynomus:
Hi @skhare-r7, I have no intention of merging this PR unfortunately. If you look at the code it is a mess. It does not follow any of the code standards that the rest of the project has.

Furthermore if you'd have looked a bit deeper into the project you see that this exact information is already gathered. Please have a look at the NtlmChallenge class. I do agree that we could for instance store that somewhere in the ConnectionInfo.

skhare-r7:
Hi @hierynomus ; no problem. I will try to re-work some of these changes more gracefully using the NTLMChallege class. I agree that the code wasn't elegant at all; what I wanted was your feedback in incorporating /exposing some of the metadata in the ConnectionInfo.

PR #189, opened 2017-09-11
Exposing WindowsVersion as a part of NTLM connection setup

skhare-r7:
Extracting and exposing metadata as a part of connection setup that is available even if the connection setup fails.

### DIFF
--- a/src/main/java/com/hierynomus/ntlm/messages/NtlmChallenge.java
+++ b/src/main/java/com/hierynomus/ntlm/messages/NtlmChallenge.java
@@ -152,4 +152,6 @@ public class NtlmChallenge extends NtlmPacket {
     public Object getAvPairObject(AvId key) {
         return this.targetInfo.get(key);
     }
+
+    public WindowsVersion getVersion() { return version; }
 }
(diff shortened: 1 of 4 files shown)


skhare-r7:
@hierynomus ; any comments on this approach?

hierynomus:
Yes, I have comments :)

I do not like this approach. You're storing the WindowsVersion all over the place. The Authenticator should not care about storing that. I feel that this is a weird thing in the protocol that you only 'discover' this kind of info in the authentication rounds, opposed to it being communicated when you negotiate the dialect.

My suggestion:
- Make `ConnectionInfo` a public class with public getters for the necessary information, but keep constructor and initialization methods package private
- Move `sessionTable`, `preauthSessionTable`, `outstandingRequests` and `sequenceWindow` to the Connection class and don't expose them. They're really `Connection` properties and should not be exposed outside of that.
- Return from the Authenticators not a plain `byte[]` but some other object which contains the `byte[]` and any additional information (like the `WindowsVersion`).
- Set the `WindowsVersion` and other additional details gathered in the Authenticator in the `ConnectionInfo` but don't store it somewhere else.
- Check whether we can get the same info from the `SpnegoAuthenticator` and return it from there too.

skhare-r7:
Thanks for the input! Ill try this approach.

skhare-r7:
Closing this PR, will open another with the suggestions.

PR #190, opened 2017-09-13
Adding windows version info to ConnectionInfo

skhare-r7:
As discussed:
1. Moving more `ConnectionInfo` oriented fields out of `Connection` 
2. Adding `WindowsVersion` field to `ConnectionInfo`
3. Returning `WindowsVersion` and the negotiated token as an `AuthenticateResponse` object, instead of `byte[]`

Currently the `AuthenticateResponse` class is bare; may make additions in the future based on useful information extracted from negotiation.


hierynomus:
Can you check that you follow the same code-style? Ie. Open-brace at end-of-line.

Also don't make fields public

skhare-r7:
Put this on hold, I'm making a few more changes that I have not pushed.

Issue #184, opened 2017-09-07
Status_Access_Denied while accessing Samba Share on Linux


Varsha-Nagarajan:
Scenario : Unable to access samba share mounted on a linux system (SUSE Linux Enterprise)
Samba server version : 3.6.3
I am using anonymous and guest authentication context but the connection to the share fails with:

com.hierynomus.mssmb2.SMBApiException: STATUS_ACCESS_DENIED(3221225506/3221225506): Could not connect to \\ipaddress\shareName
	at com.hierynomus.smbj.session.Session.connectTree(Session.java:113)
	at com.hierynomus.smbj.session.Session.connectShare(Session.java:98)
	at TestIllegalArgument.main(TestIllegalArgument.java:52)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:147)

The smbd log file says:

[2017/09/07 20:47:41.270463,  0] ../lib/util/util.c:415(dump_data)
  [0000] 61 EB 0C 0E A9 95 C4 E9   0F 17 AE 8D 46 7E F4 36   a....... ....F~.6
[2017/09/07 20:47:41.270587,  0] ../lib/util/util.c:415(dump_data)
  [0000] 69 43 23 A7 4B 21 28 80   F8 E4 C8 8F 89 83 A4 4A   iC#.K!(. .......J
[2017/09/07 20:47:53.799473,  0] smbd/smb2_signing.c:129(smb2_signing_check_pdu)
  Bad SMB2 signature for message

The SMB conf file looks like:

[content-repo]
        browseable = Yes
        path = path to share
        public = Yes
        writeable = Yes
        read only = Yes
        read list = nobody guest
        valid users = nobody guest
        create mask = 0664

Please tell me whether there is anything else that needs to be configured. I verified that the share permits SMBv2 connections. Both anonymous and guest authentication keeps failing. The folder being shared has all permissions and is being owned by "nobody".

I have tried adding guest account = nobody, guest ok = yes and force user = nobody. Still no luck.

hierynomus:
Which version of smbj are you using?

[...]

PR #199, opened 2017-09-28
Correctly disable signing in case of guest/anonymous authentication

### LINKED ISSUE #196: Disabling signatures when guest/anonymous
We've had an ongoing discussion here around guest sessions: https://github.com/hierynomus/smbj/issues/184#issuecomment-329306762

The conclusion I've come to is that SMB does not allow for signing when the session is resolved to be a guest or anonymous: https://blogs.msdn.microsoft.com/openspecification/2017/05/26/smb-2-and-smb-3-security-in-windows-10-the-anatomy-of-signing-and-cryptographic-keys/
>
> To sign a message, the first two prerequisites are:
> • Existence of the signing key: SessionId field has a non-zero value, otherwise there would be no possible signing key. The exception is a guest or anonymous session, which does not have a signing key even there’s a SessionId.
> • Signing requirement: SessionId requires signing, i.e. its Session.SigningRequired is equal to TRUE as outlined previously.

Since from the perspective of an smbj user, I won't know that the session is guest or anonymous until the `Session` is returned to me. I've supported `SmbConfig.withSigningRequired` by checking if the returned Session is a guest or anonymous after it is returned, disabling signing if it is, and then reconnecting.

I feel like smbj could support this without forcing the reconnect from the user, perhaps by checking after it has completed the `SMB2_SESSION_SETUP` request and choosing not to calculate a secret key.  What do you think @hierynomus ?

hierynomus:
Fixes #196 and #184 

@Varsha-Nagarajan and @tosmun-r7 Can you verify whether this indeed fixes the problem?

### DIFF
--- a/src/main/java/com/hierynomus/smbj/session/Session.java
+++ b/src/main/java/com/hierynomus/smbj/session/Session.java
@@ -95,7 +95,7 @@ public class Session implements AutoCloseable {
             signingRequired = false;
         }
         if (guest && connection.getConfig().isSigningRequired()) {
-            throw new SMBRuntimeException("Cannot require message signing when authenticating with a guest account");
+            throw new SMB2GuestSigningRequiredException();
         } else if (guest) {
             signingRequired = false;
         }
--- a/src/main/java/com/hierynomus/smbj/session/Session.java
+++ b/src/main/java/com/hierynomus/smbj/session/Session.java
@@ -95,7 +95,7 @@ public class Session implements AutoCloseable {
             signingRequired = false;
         }
         if (guest && connection.getConfig().isSigningRequired()) {
-            throw new IllegalStateException("Cannot require message signing when authenticating with a guest account");
+            throw new SMBRuntimeException("Cannot require message signing when authenticating with a guest account");
         } else if (guest) {
             signingRequired = false;
         }
--- a/src/main/java/com/hierynomus/smbj/session/Session.java
+++ b/src/main/java/com/hierynomus/smbj/session/Session.java
@@ -46,24 +46,24 @@ public class Session implements AutoCloseable {
     private long sessionId;
 
     private PacketSignatory packetSignatory;
-    private boolean serverSigningRequired;
+    private boolean signingRequired;
+    private boolean encryptData; // SMB3.x
 
     private Connection connection;
     private SMBEventBus bus;
     private boolean dfsEnabled;
     private TreeConnectTable treeConnectTable = new TreeConnectTable();
-    private AuthenticationContext auth;
+    private AuthenticationContext userCredentials;
 
     private boolean guest;
     private boolean anonymous;
 
-    public Session(Connection connection, AuthenticationContext auth, SMBEventBus bus, boolean signingRequired, boolean dfsEnabled, SecurityProvider securityProvider) {
+    public Session(Connection connection, AuthenticationContext userCredentials, SMBEventBus bus, boolean dfsEnabled, SecurityProvider securityProvider) {
         this.connection = connection;
-        this.auth = auth;
+        this.userCredentials = userCredentials;
         this.bus = bus;
         this.dfsEnabled = dfsEnabled;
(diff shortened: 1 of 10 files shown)


src/main/java/com/hierynomus/smbj/session/Session.java
```diff
@@ -72,6 +72,37 @@ public Session(Connection connection, AuthenticationContext auth, SMBEventBus bu
...
+        if (requireMessageSigning || connectionSigningRequired) {
+            signingRequired = true;
+        }
+
+        if (anonymous) {
+            // If the security subsystem indicates that the session was established by an anonymous user, Session.SigningRequired MUST be set to FALSE.
+            signingRequired = false;
+        }
+        if (guest && connection.getConfig().isSigningRequired()) {
+            throw new IllegalStateException("Cannot require message signing when authenticating with a guest account");
```
tosmun-r7: Could we use a specific `SmbRuntimeException` for this?

hierynomus: Yes, will do

tosmun-r7:
Another question @hierynomus . Would it be possible to disable the signing requirement from the client but maintain the original connection? Since it's per-session. It would probably require the client config to be mutable though.

hierynomus:
Not sure what you mean. The implementation now is entirely according to the spec. The Config object maintains the global RequireMessageSigning flag. So if that is set and you're guest-authenticating, it will throw an exception, otherwise your connection will succeed

tosmun-r7:
Ah I see, so we'll treat all sessions on all connections produced by the client to require signing. Rather than a per-session basis.

tosmun-r7:
Ah I see. If we request signing in the session (without knowing it's a guest ahead of time for example), and it ends up being guest, the spec states to close the connection. Got it.
> If the SMB2_SESSION_FLAG_IS_GUEST bit is set in the SessionFlags field of the SMB2 SESSION_SETUP Response and if Session.SigningRequired is TRUE, this indicates a SESSION_SETUP failure and the connection MUST be terminated.

hierynomus:
Have you experimented with the implementation already @tosmun-r7 ? If so, I'll merge this

tosmun-r7:
I'll test it out today and get back to you, thanks!

src/main/java/com/hierynomus/smbj/session/Session.java
```diff
@@ -72,6 +72,37 @@ public Session(Connection connection, AuthenticationContext auth, SMBEventBus bu
...
+        if (requireMessageSigning || connectionSigningRequired) {
+            signingRequired = true;
+        }
+
+        if (anonymous) {
+            // If the security subsystem indicates that the session was established by an anonymous user, Session.SigningRequired MUST be set to FALSE.
+            signingRequired = false;
+        }
+        if (guest && connection.getConfig().isSigningRequired()) {
+            throw new SMBRuntimeException("Cannot require message signing when authenticating with a guest account");
```
tosmun-r7: Could we use a really specific exception for this case that extends `SMBRuntimeException`? It will help with use cases where the client may wish to open a new connection without signing should they discover it's a guest account. i.e. `SMBGuestSigningViolationException` or something?

tosmun-r7: Specifically the logic I'd like to capture is:
```java
Session session = null;
try {
   session = conn.authenticate(auth);
} catch (SMBGuestSigningViolationException e) {
   session.close();
   conn.close();
   conn = // No signing this time, and try again
}
```

The reason for this is I'd like to enforce signing in all cases *except* for guest, where it is not applicable.

hierynomus:
@tosmun-r7 Anything more? Were you able to run it through some more tests?

tosmun-r7:
@hierynomus Tested it just now and it works great, thanks! 🚢

Varsha-Nagarajan:
@hierynomus sorry for the late reply. I tested it too and it works just great. I really appreciate it. Thanks :)

Issue #198, opened 2017-09-28
Signature verification for CHANGE_NOTIFY fails


ndimitry:
All other tested requests seem to work fine (open, close, list, delete, read and write).
As for the watch it returns correct contents but produces warnings which state that signature verification fails. Here's a warning example that I get:
ERROR | Signatures for packet SMB2_CHANGE_NOTIFY with message id << 504251 >> do not match (received: [-44, 12, -16, -51, 46, -47, -124, -120, -67, -72, 92, 44, 31, -120, 71, -61], calculated: [-29, -122, -92, 11, -109, 50, -11, 104, 45, -118, 126, -20, -48, -119, -100, 109, -30, -85, 77, 44, 41, 114, -103, -22, -114, -91, 18, -76, -124, -5, 37, -33])

Connection#verifyPacketSignature fails since a calculated and returned signatures are different. Either I do something wrong, either there is some kind of a defect there.

Here is a peace of code that faces this issue.

```
    void smbjWatch() throws Exception {
        SMBClient client = new SMBClient();

        try (Connection connection = client.connect("host")) {
            AuthenticationContext ac = new AuthenticationContext("user", "password".toCharArray(), "");
            Session session = connection.authenticate(ac);

            // Connect to Share
            try (DiskShare share = (DiskShare) session.connectShare("share")) {
                Set<AccessMask> accessMasks = new HashSet<>();
                accessMasks.add(AccessMask.FILE_LIST_DIRECTORY);
                // Open the root share folder
                try (Directory directory = share.openDirectory("", accessMasks, null,
                        SMB2ShareAccess.ALL, SMB2CreateDisposition.FILE_OPEN, null)) {
                    // Track all changes until program is terminated
                    while (true) {
                        Set<SMB2CompletionFilter> completionFilter = new HashSet<>();
                        completionFilter.add(SMB2CompletionFilter.FILE_NOTIFY_CHANGE_FILE_NAME);
                        SMB2ChangeNotifyRequest request = new SMB2ChangeNotifyRequest(
                                session.getConnection().getNegotiatedProtocol().getDialect(),
                                session.getSessionId(),
                                share.getTreeConnect().getTreeId(),
                                directory.getFileId(),
                                completionFilter,
                                65536,
                                true);

                        Future<SMB2ChangeNotifyResponse> future = session.send(request);
                        SMB2ChangeNotifyResponse response = future.get();
                        // just some output to make life easier
[...]

PR #200, opened 2017-09-28
Fix for issue #198, signature verification fails for CHANGE_NOTIFY response

ndimitry:
When there are 2 or more actions returned the buffer read position is calculated incorrectly.


src/main/java/com/hierynomus/mssmb2/messages/SMB2ChangeNotifyResponse.java
```diff
@@ -72,8 +72,12 @@ protected boolean isSuccess(NtStatus status) {
             fileNameLen = buffer.readUInt32();
             fileName = buffer.readString(StandardCharsets.UTF_16LE, (int) fileNameLen / 2);
             notifyInfoList.add(new FileNotifyInfo(action, fileName));
-            currentPos += nextEntryOffset;
-            buffer.rpos(currentPos);
+            if (nextEntryOffset != 0) {
+                currentPos += nextEntryOffset;
+                buffer.rpos(currentPos);
+            } else {
```
hierynomus: There is no need for the else. After reading the last fileName the buffer is at the right position.

Issue #142, opened 2017-08-03
Print question


gidravlic:
Hello! In the old version of the library(0.9) i could print through method:
```java
static void write(PrinterShare share, String remotePath, InputStream is) throws IOException, SMBApiException {
        SMB2FileId fileId = share.open(
                remotePath,
                AccessMask.FILE_WRITE_DATA.getValue(),
                EnumSet.of(FileAttributes.FILE_ATTRIBUTE_NORMAL),
                EnumSet.of(SMB2ShareAccess.FILE_SHARE_WRITE),
                SMB2CreateDisposition.FILE_CREATE,
                EnumSet.of(SMB2CreateOptions.FILE_WRITE_THROUGH));
        com.hierynomus.smbj.share.File file = new File(fileId, share.getTreeConnect(), remotePath, 0L);
        file.write(new InputStreamByteChunkProvider(is));
        file.close();
}
```
With latest version of the library, this code don't work. 
How to print to printer right now?
Thanks!

gidravlic:
Hi @hierynomus !
After searching problem, i'm discovered that File class constructor takes only DiskShare parameter. I changed PrinterShare class, instead:
```java 
public class PrinterShare extends Share 
```
I extend PrinterShare from DiskShare
```java
public class PrinterShare extends DiskShare {

    public PrinterShare(SmbPath smbPath, TreeConnect treeConnect) {
        super(smbPath, treeConnect);
    }

    public void print(String fileName, InputStream content) {
        try (File file = openFile(
                fileName,
                EnumSet.of(AccessMask.FILE_WRITE_DATA),
                EnumSet.of(FileAttributes.FILE_ATTRIBUTE_NORMAL),
                EnumSet.of(SMB2ShareAccess.FILE_SHARE_WRITE),
                SMB2CreateDisposition.FILE_CREATE,
                EnumSet.of(SMB2CreateOptions.FILE_WRITE_THROUGH))) {
            file.write(new InputStreamByteChunkProvider(content));
        }
    }
}
```
calling work perfectly
```java
printerShare.print("test.txt", new ByteArrayInputStream("Test Test Test".getBytes(StandardCharsets.UTF_8)));
```
I do not quite understand the whole architecture of the library in order to make appropriate changes to it. What thoughts?

Thanks!

pepijnve:
[...]

PR #201, opened 2017-09-29
#142 Restore print functionality

pepijnve:


hierynomus:
You beat me to it @pepijnve See #202 :)
If you don't mind I'll merge that one, as I also did a small refactoring there 😉

pepijnve:
I don't mind at all. I had considered the same 'common writer' refactoring, but ended up trying to keep the change as minimal as possible.

PR #205, opened 2017-10-02
Avoid returning a closed connection by the SMBClient class

ndimitry:
Currently SMBClient#connect can return a broken Connection.

```
    private Connection getEstablishedOrConnect(String hostname, int port) throws IOException {
        synchronized (this) {
            String hostPort = hostname + ":" + port;
            if (!connectionTable.containsKey(hostPort)) {
                // create a connection, put it to the connectionTable and return it
            }
            // if a connection is broken/closed we still returning it
            return connectionTable.get(hostPort);
        }
    }
```


ndimitry:
My previous commits are also exposed in this pull request, maybe I've made something incorrectly..

ndimitry:
Reason for this pull request is the fact that sometimes because of exceptions or other reasons socket gets closed but still this broken connection is returned by the client

ndimitry:
@hierynomus please, review (does this pull request makes sense or do I miss something?)

PR #179, opened 2017-09-05
throw error when state is not DFSPath

cliviu:

PR #218, opened 2017-10-15
ReferralCache fixes

pepijnve:
- Implement ReferralCache#lookup
- Correctly set interlink flag in single storage server response case
- Correct ReferralCacheEntry#isExpired implementation


pepijnve:
@hierynomus I think this fix makes PR #179 obsolete. While going through the DFS path resolution code I did not encounter any other errors in the implementation except for that one TODO which has now been implemented.

PR #219, opened 2017-10-15
ReferralCache#lookup should find deepest match instead of exact match

pepijnve:


pepijnve:
Thinking about this change some more. I still didn't get it right since now it only checks prefixes. That means an entry for `/abc/de` would also be returned for a lookup for `/abc/def`. The lookup should actually only look at complete path components. Maybe I should just go ahead and implement a basic trie for this after all.
For the time being it might be better to revert lookup to `return null`.

hierynomus:
I was already doing that ;)

hierynomus:
ie. implementing the trie ;)

PR #220, opened 2017-10-16
Using trie to store referral cache entries

hierynomus:
@pepijnve Please have a look to see whether I missed something.

### DIFF
--- a/src/main/java/com/hierynomus/msdfsc/ReferralCache.java
+++ b/src/main/java/com/hierynomus/msdfsc/ReferralCache.java
@@ -19,10 +19,7 @@ import com.hierynomus.msdfsc.messages.DFSReferral;
 import com.hierynomus.msdfsc.messages.SMB2GetDFSReferralResponse;
 import com.hierynomus.msdfsc.messages.SMB2GetDFSReferralResponse.ReferralHeaderFlags;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.*;
 
 /**
  * [MS-DFSC].pdf: 3.1.1 Abstract Data Model
@@ -61,7 +58,7 @@ import java.util.concurrent.ConcurrentHashMap;
  */
 public class ReferralCache {
 
-    private ConcurrentHashMap<String, ReferralCacheEntry> cache = new ConcurrentHashMap<>();
+    private ReferralCacheNode cacheRoot = new ReferralCacheNode("<root>");
 
     public static class TargetSetEntry {
         String targetPath;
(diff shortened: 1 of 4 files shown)


src/main/java/com/hierynomus/msdfsc/ReferralCache.java
```diff
@@ -183,4 +170,46 @@ public String toString() {
 
     }
 
+
+    static class ReferralCacheNode {
+        String pathComponent;
+        Map<String, ReferralCacheNode> childNodes = new HashMap<>();
```
pepijnve: Use ConcurrentHashMap to avoid potential ConcurrentModificationexceptions?

src/main/java/com/hierynomus/msdfsc/ReferralCache.java
```diff
@@ -183,4 +170,46 @@ public String toString() {
...
+
+        ReferralCacheNode(String pathComponent, ReferralCacheEntry entry) {
+            this.pathComponent = pathComponent;
+            this.entry = entry;
+        }
+
+        void addReferralEntry(Iterator<String> pathComponents, ReferralCacheEntry entry) {
+            if (pathComponents.hasNext()) {
+                String component = pathComponents.next();
+                if (!childNodes.containsKey(component)) {
```
pepijnve:
```
ReferralCacheNode node = childNodes.get(component);
if (node == null) {
    node = new ReferralCacheNode(component);
    childNodes.put(component, node);
}
node.addReferralEntry(pathComponents, entry);
```

That'll be slightly faster. `contains` followed by `get` means you effectively do the same hash table lookup twice.

src/main/java/com/hierynomus/msdfsc/ReferralCache.java
```diff
@@ -183,4 +170,46 @@ public String toString() {
...
+                childNodes.get(component).addReferralEntry(pathComponents, entry);
+            } else {
+                this.entry = entry;
+            }
+        }
+
+        ReferralCacheEntry getReferralEntry(Iterator<String> pathComponents) {
+            if (pathComponents.hasNext()) {
+                String component = pathComponents.next();
+                if (childNodes.containsKey(component)) {
```
pepijnve: See comment on addReferralEntry

src/main/java/com/hierynomus/msdfsc/ReferralCache.java
```diff
@@ -183,4 +170,46 @@ public String toString() {
...
+        }
+
+        ReferralCacheNode(String pathComponent, ReferralCacheEntry entry) {
+            this.pathComponent = pathComponent;
+            this.entry = entry;
+        }
+
+        void addReferralEntry(Iterator<String> pathComponents, ReferralCacheEntry entry) {
+            if (pathComponents.hasNext()) {
+                String component = pathComponents.next();
```
pepijnve: Should case normalisation (e.g., toLowerCase) be done here and in getReferralEntry to get case insensitive lookups? In general SMB is case preserving, so case sensitive caching might be a bit stricter than necessary. On the other hand it doesn't cause correctness issues, only potential redundant referral requests.

src/main/java/com/hierynomus/msdfsc/ReferralCache.java
hierynomus: Yes, indeed.

src/main/java/com/hierynomus/msdfsc/ReferralCache.java
hierynomus: What does the spec say? As far as I can see only the server is required to treat paths in a case-insensitive manner, nothing is said about the cache.

pepijnve: Doesn't mention it afaik.

PR #230, opened 2017-11-01
prevent NPE when use close() twice or more

xuthus:
The behavior of class com.hierynomus.smbj.share.FileOutputStream differs from the behavior of java.io.FileOutputStream when I call close() method twice.

java.io implementation allows to close() already closed object, smbj implementation, I think, allows by design, but prevents by bug.

Of course, it is not good practice to close stream twise, but when we use withCloseable() and closure closes PrintWriter using the stream, its close() called twice.

Correct me, please, if I'm wrong.

### DIFF
--- a/src/main/java/com/hierynomus/smbj/share/FileOutputStream.java
+++ b/src/main/java/com/hierynomus/smbj/share/FileOutputStream.java
@@ -119,7 +119,7 @@ class FileOutputStream extends OutputStream {
 
         @Override
         public boolean isAvailable() {
-            return !buf.isEmpty();
+            return buf != null && !buf.isEmpty();
         }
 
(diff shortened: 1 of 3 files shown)


hierynomus:
Change looks perfect, could you add a test to the unit tests that prevents regression?

xuthus:
Added 2 tests. Please, review its, may be I used BasicPacketProcessor or requests handling incorrectly. Anyway, those tests cover changed code, but its can fail in future in case of wrong realisation.

p.s. Spock is easier than SMB :)

hierynomus:
Pretty extensive test :D I would've been happy with a mocked out SmbWriter ;).
Thanks!

PR #238, opened 2017-11-10
Make Connection implement Closeable instead of AutoCloseable (and throw IOException on close)

ndimitry:


hierynomus:
@ndimitry Why this change?
`AutoCloseable.close()` specifies that the exception coming out can be any type. The old `Closeable` interface narrows this to `IOException`. However we implement `Autocloseable` here.

ndimitry:
I think in AutoClosable it is done to provide the possibility to use try-with-resources in any custom case (AutoClosable provides the most general case). By implementing it one may narrow Exception to some specific exception classes.

For example in our product we use SmpOperationException (which is a wrapper over SmpApiException) and we have:
`abstract class SmbItemWrapper<T extends DiskEntry> implements AutoCloseable {
    @Override
    public void close() throws SmbOperationException {
        // do some stuff
    }
}`
In this case I will be forced to catch only SmbOperationException when using try-with-resources.

And the main reason for this suggestion is the fact that when we close the connection we are forced to catch Exception (everything), which I guess is not the case. Only IOExceptions may be expected as the result of Connection#close() (as far as I see). On the other hand by catching Exception we are forced to catch, for example, NullPointerException which extends Exception and RuntimeException and thus is not expected to be caught.

ndimitry:
Similar thoughts in the reply to this question https://stackoverflow.com/questions/25959779/why-close-method-of-java-lang-autocloseable-throws-exception-but-close-method-o

hierynomus:
In that case it might make more sense to just implement `Closeable`. Can you change the PR to reflect that?

ndimitry:
Changed it co Closeable (and pushed a new commit to implement that).

ndimitry:
Thank you for being open for the changes and fast to react!
Both are valuable!

hierynomus:
You're welcome. Thanks for the useful contributions.

PR #248, opened 2017-11-16
With intigua changes 2

noaml2gitHub:


hierynomus:
@noaml2gitHub What's this?
A. It has serious conflicts with master
B. It seems that all of these changes are already present

noaml2gitHub:
You are right.
My bad...

Issue #254, opened 2017-11-21
Could not connect to Share folder in DFS 


harisingh-highq:
Hi @hierynomus
I have tried with below code as sample:-
SmbConfig config = SmbConfig.builder()
			    .withMultiProtocolNegotiate(true)
			    .withSigningRequired(true)
			    .withDfsEnabled(true)
			    .build();
			SMBClient client = new SMBClient(config);
			Connection conn = client.connect("domain name");
			Session session = conn.authenticate(new AuthenticationContext("hari.singh", "hari".toCharArray(), ""));
			DiskShare share = (DiskShare) session.connectShare("data\\abc.xyz");

The below exceptions occurs after above code execution:-
com.hierynomus.mssmb2.SMBApiException: STATUS_BAD_NETWORK_NAME(3221225676/3221225676): Could not connect to \\domain name\data\haresh.bhatt
	at com.hierynomus.smbj.session.Session.connectTree(Session.java:147)
	at com.hierynomus.smbj.session.Session.connectShare(Session.java:132)
	at TestSMBJ_DFS.dfsConnection(TestSMBJ_DFS.java:66)
	at TestSMBJ_DFS.main(TestSMBJ_DFS.java:37)

With same DFS server configuration, i have tried with JCIFS library and it succeed with DFS which ensures that DFS environment is correct but it could not resolve DFS properly with SMBJ library

Here, if i try with withDfsEnabled() property set to false in SMB client then also it throws same error.

Can you please help me to sort out this issue?

pepijnve:
This is most likely the same issue as #212. Possible fix in PR #222.

hierynomus:
Also a problem could be that you're trying to connect to this share:

`DiskShare share = (DiskShare) session.connectShare("data\abc.xyz");`

However a share name does not contain a `\`. The share is `data`, and then you can open a file/folder called `abc.xyz`. If it is a DFS share, it will then transitively resolved to the correct directory.

harisingh-highq:
Hi @hierynomus 
I have already tried with data share folder only but its not working fine.
In our infrastructure, we have data folder under domain name and abc.xyz folder is shared in DFS like this domain name\\data\\abc.xyz.

hierynomus:
Yes, but from SMB perspective, the DFS is transparent. This means that abc.xyz is treated as a folder/directory, not as a share.

What happens when you split that up? What is the exception and logging you're getting then?

harisingh-highq:
Hi @pepijnve 
[...]

PR #256, opened 2017-11-22
Support for Root Referrals and cleanup of DFS

hierynomus:
This is a replacement for #222 and builds on top of that.


hierynomus:
@pepijnve Can you have a look?

src/main/java/com/hierynomus/smbj/session/Session.java
```diff
@@ -179,10 +182,10 @@ private Share connectTree(final String shareName) {
             TreeConnect treeConnect = new TreeConnect(treeId, smbPath, this, response.getCapabilities(), connection, bus);
 
             Share share;
-            if (response.isDiskShare() && dfsPathResolver != null && response.getCapabilities().contains(SMB2ShareCapabilities.SMB2_SHARE_CAP_DFS)) {
-                share = new DFSDiskShare(smbPath, treeConnect, dfsPathResolver);
-            } else if (response.isDiskShare()) {
-                share = new DiskShare(smbPath, treeConnect);
+/*            if (response.isDiskShare() && pathResolver != null && response.getCapabilities().contains(SMB2ShareCapabilities.SMB2_SHARE_CAP_DFS)) {
```
pepijnve: With this change, the library will always attempt to perform DFS referral requests even though the share may not support it. Everything can be solved with an extra level of indirection. Perhaps pathResolver should be a `Function<Set<Smb2ShareCapabilities>, PathResolver>` instead so that selection of the actual implementation can be deferred until this point.
In other words, pass a PathResolver factory to the Session instead of a concrete PathResolver. Depending on the capabilities, the factory would either return the DFSPathResolver decorator or the wrapped SymlinkPathResolver directly.

pepijnve: That being said DFSPathResolver only takes specific actions when PATH_NOT_COVERED is returned. You're unlikely to get that error unless DFS is enabled to begin with. Does the spec say anything about dealing with PATH_NOT_COVERED when SMB2_SHARE_CAP_DFS is not set?

src/main/java/com/hierynomus/smbj/paths/SymlinkPathResolver.java
```diff
@@ -0,0 +1,127 @@
...
+            String target = resolveSymlinkTarget(smbPath.getPath(), symlinkData);
+            return new SmbPath(smbPath.getHostname(), smbPath.getShareName(), target);
+        }
+
+        return wrapped.resolve(session, responsePacket, smbPath);
+    }
+
+    @Override
+    public Set<NtStatus> handledStates() {
+        return states;
```
pepijnve: This should probably be a defensive copy or be wrapped in an unmodifiable collection. Otherwise the internal state of the PathResolver is mutable.

src/main/java/com/hierynomus/smbj/paths/DFSPathResolver.java
```diff
@@ -22,37 +22,67 @@
...
+            SmbPath target = SmbPath.parse(resolve(session, smbPath.toUncPath()));
+            logger.info("DFS resolved {} -> {}", smbPath, target);
+            return target;
+        }
+        return wrapped.resolve(session, responsePacket, smbPath);
+    }
+
+    @Override
+    public Set<NtStatus> handledStates() {
+        return this.states;
```
pepijnve: This should probably be a defensive copy or be wrapped in an unmodifiable collection. Otherwise the internal state of the PathResolver is mutable.

hierynomus:
@pepijnve Can you check whether this now exhibits the same behaviour as your original commit?

harisingh-highq:
@hierynomus  
May i know that what is the expected time for DFS working SMBJ jar. As we need to work on our side for production roll out. 
For reference #254

hierynomus:
@harisingh-highq Can you try and test this PR? If it works, I'll merge it so that it's ready for a new release.

hierynomus:
@pepijnve Do you have any time to validate this PR on the root referral scenario? I don't have a setup for that ...

pepijnve:
I haven't had a chance to try it out yet. I might be able to do some quick tests later today.

hierynomus:
👍 Would be great! Thanks!

pepijnve:
Remote DFS namespaces didn't work correctly. See PR #258.

hierynomus:
@pepijnve given the comments on #258 it is good to merge?

pepijnve:
Yep, root referrals work for me. The only limitation that's still left that I know of is that the DFS client only uses the TargetHint at the moment and ignores the TargetList. See last comment on #254. I think that's out of scope for this PR though.

hierynomus:
Yes, let's tackle that in a different PR :)

Disablez:
Definitely a step forward; fixed some issues for me as well. Nice job!

PR #275, opened 2018-01-10
Implementing the writeAsync method for file and share

yin19941005:
Implement the writeAsync method in order to send the write request easier from a file perceptive. ArrayList is used to handle the case multiple write requests are required to send (i.e. multiple write response future should return). Without this commit, we can still got sufficient information from the diskShare to send a write request through the corresponding session. This commit just aim to make it simpler.


src/main/java/com/hierynomus/smbj/share/File.java
```diff
@@ -87,6 +89,29 @@ public int write(ByteChunkProvider provider, ProgressListener progressListener)
...
 
+    /***
+     * Write the data Async in buffer to this file at position fileOffset.
+     * @param buffer     the data to write
+     * @param fileOffset The offset, in bytes, into the file to which the data should be written
+     * @param offset     the start offset in the data
+     * @param length     the number of bytes that are written
+     * @return the Array List of write response future
+     */
+    public ArrayList<Future<SMB2WriteResponse>> writeAsync(byte[] buffer, long fileOffset, int offset, int length) {
```
hierynomus: Please make the return type `List<...>` instead of `ArrayList<...>` (and everywhere else in this PR

pepijnve:
I think there's a bit of an issue with this patch (and in SMBJ itself already today) when performing multiple consecutive writes. Currently the code assumes that if I request a write of `n` bytes from the server, that `n` bytes will actually have been written. This isn't guaranteed to be the case though, the server might return a write result that says fewer than `n` bytes have been written.
SMBJ ignores this though since it optimistically consumes the `n` bytes from the `ByteChunkProvider` when it sends out the write request and doesn't rewind if fewer bytes were written.

Dealing with this case in the synchronous case is already going to be a bit of a pain. Handling it in the async case is going to be even trickier.

@hierynomus should I log an issue for this? Or is this only a theoretical possibility that never happens in practice?

yin19941005:
Hi @pepijnve ,

Yes, I think the case mentioned about the expected written bytes mismatch the actual written bytes can be happen and the write provider (e.g. the ArrayByteChunkProvider) doesn't seems handling this case (if the status is success status). But I doubt this shouldn't happens, if a write request is not completed in whole, will the server simply reply `STATUS_SUCCESS`?

I just checked the `3.3.5.13 Receiving an SMB2 WRITE Request` on [MS-SMB2], it states : 

> The status code returned by this operation MUST be one of those defined in [MS-ERREF]. Common status codes returned by this operation include:
>  STATUS_SUCCESS
>  STATUS_INSUFFICIENT_RESOURCES
>  STATUS_ACCESS_DENIED
>  STATUS_FILE_CLOSED
>  STATUS_NETWORK_NAME_DELETED
>  STATUS_USER_SESSION_DELETED
>  STATUS_NETWORK_SESSION_EXPIRED
>  STATUS_INVALID_PARAMETER
>  STATUS_PIPE_BROKEN
>  STATUS_DISK_FULL
>  STATUS_CANCELLED
>  STATUS_FILE_LOCK_CONFLICT

In my opinion, these status are not able to represent the situation that expected written bytes mismatch the actual written bytes. What do you think about it?

pepijnve:
The mismatch can be detected by comparing [the length field from the write request](https://msdn.microsoft.com/en-us/library/cc246532.aspx) with [the count field from the write result msg](https://msdn.microsoft.com/en-us/library/cc246533.aspx).

[SMB2Writer](https://github.com/hierynomus/smbj/blob/master/src/main/java/com/hierynomus/smbj/share/SMB2Writer.java#L92) already uses the value from the write result, but ignores the fact that it may be smaller than the number of bytes that were consumed from the ByteChunkProvider in [SMB2WriteRequest#writeTo](https://github.com/hierynomus/smbj/blob/master/src/main/java/com/hierynomus/mssmb2/messages/SMB2WriteRequest.java#L51). Looking at that code in particular; it's actually a bit of a poor design choice that [SMB2WriteRequest#writeTo](https://github.com/hierynomus/smbj/blob/master/src/main/java/com/hierynomus/mssmb2/messages/SMB2WriteRequest.java#L39) has the extra side effect of modifying the ByteChunkProvider. Anyway, I'm going off topic. I'll make a separate issue/pr to clean up that part of the code.

yin19941005:
Yes, I understand you mean it can be "detected" but I am doubting will the mismatch possible.

If yes (the mismatch case is possible), then we shouldn't update the bufferOffset in [ArrayByteChunkProvider#getChunk](https://github.com/hierynomus/smbj/blob/master/src/main/java/com/hierynomus/smbj/io/ArrayByteChunkProvider.java#L49) automatically.

I believe this is a related topic because if this issue exist, the writeAsync method would not guarantee the result as well. However, I agree that we should make a a separate issue/pr to discuss or solve this issue.

hierynomus:
As far as I can see in the spec, if the status is `SUCCESS`, the number of bytes written is what was requested.

```
3.3.5.13 Receiving an SMB2 WRITE Request
[...]
If Open.IsSharedVHDX is FALSE, the server MUST issue a write to the underlying object store represented by Open.LocalOpen for the length, in bytes, given by Length, at the offset, in bytes, from the beginning of the file, provided in Offset. 
```

There is no mention in the server specifics that less bytes could be written. 

Also if it were possible that the server would return success without writing everything, it would not be easily possible to send out async writes, which according to the spec is possible.

hierynomus:
So I think this is not only highly theoretical, but also improbable ;)

pepijnve:
> I am doubting will the mismatch possible.

I don't know either. It could happen in theory but might never actually happen in practice. As far as I can tell the specification does not require that they are the same. A robust client library should not make the assumption that they are. At the very least, if a mismatch is detected an error should be reported since otherwise data corruption is happening silently (i.e., you think all your data was written while it has not).

hierynomus:
I can live with that, throwing an error if the response data does not match the sent data. Once we see that occurring we can tackle the problem...

hierynomus:
@pepijnve @yin19941005 I've (finally) tackled this PR, can you have a look at the latest commit which does what we discussed above.

hierynomus (approved):

Issue #276, opened 2018-01-10
Question: Is FILE_READ_DATA flag necessary on diskShare.getFileInformation()?


yin19941005:
Hi all,

At this line: 
[https://github.com/hierynomus/smbj/blob/ea13ee9fe93884a967d8a085bc7ead3443d075a0/src/main/java/com/hierynomus/smbj/share/DiskShare.java#L242](https://github.com/hierynomus/smbj/blob/ea13ee9fe93884a967d8a085bc7ead3443d075a0/src/main/java/com/hierynomus/smbj/share/DiskShare.java#L242)

We are using `GENERIC_READ` flag which is combination of access flags listed above: `FILE_READ_DATA| FILE_READ_ATTRIBUTES| FILE_READ_EA| SYNCHRONIZE| READ_CONTROL`

I tried to use flags combination `FILE_READ_ATTRIBUTES| FILE_READ_EA| SYNCHRONIZE| READ_CONTROL` to get the `FileAllInformation` and it had succeed. I am wondering is there any file information is requiring this flag to got the file information. I am concerning this one is because this would cause `STATUS_SHARING_VIOLATION` if we mis-setting the `FILE_READ_DATA` flag on.

pepijnve:
In order to query `FILE_ALL_INFORMATION` I think `FILE_READ_ATTRIBUTES | FILE_READ_EA` is actually sufficient. `SYNCHRONIZE` is not relevant and `READ_CONTROL` is only necessary when querying the DACL of the file/directory. As you stated `FILE_READ_DATA` is only required if the contents of a file or directory are going to be read which is not the case here either.

yin19941005:
Hi @pepijnve ,

Thank you for your reply! I see your point. Now, I think always using `FILE_READ_ATTRIBUTES | FILE_READ_EA` for the method `public <F extends FileQueryableInformation> F getFileInformation(String path, Class<F> informationClass) {}` is sufficient.

And for the getSecurityInfo method `public SecurityDescriptor getSecurityInfo(String path, Set<SecurityInformation> securityInfo) throws SMBApiException {}` on this line:
[...]

PR #277, opened 2018-01-10
Avoid excessive privilege on set/query file info/security info

yin19941005:
Originally, we are directly using GENERIC_READ/GENERIC_WRITE and this would cause excessive privilege. This commit is avoiding the excessive privilege issue on set or query for both file information class and the security information class.

This is related with #276 .

### DIFF
--- a/src/main/java/com/hierynomus/smbj/share/DiskShare.java
+++ b/src/main/java/com/hierynomus/smbj/share/DiskShare.java
@@ -209,7 +209,7 @@ public class DiskShare extends Share {
      * @see Directory#iterator(Class, String)
      */
     public <I extends FileDirectoryQueryableInformation> List<I> list(String path, Class<I> informationClass, String searchPattern) {
-        try (Directory d = openDirectory(path, of(GENERIC_READ), null, ALL, FILE_OPEN, null)) {
+        try (Directory d = openDirectory(path, of(FILE_LIST_DIRECTORY, FILE_READ_ATTRIBUTES, FILE_READ_EA), null, ALL, FILE_OPEN, null)) {
             return d.list(informationClass, searchPattern);
         }
     }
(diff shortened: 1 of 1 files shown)


src/main/java/com/hierynomus/smbj/share/DiskShare.java
```diff
@@ -398,7 +398,7 @@ public SecurityDescriptor getSecurityInfo(SMB2FileId fileId, Set<SecurityInforma
      * The SecurityDescriptor(MS-DTYP 2.4.6 SECURITY_DESCRIPTOR) for the Given FileId
      */
     public void setSecurityInfo(String path, Set<SecurityInformation> securityInfo, SecurityDescriptor securityDescriptor) throws SMBApiException {
-        Set<AccessMask> accessMask = of(GENERIC_WRITE);
+        Set<AccessMask> accessMask = of(READ_CONTROL);
```
hierynomus: I would assume we would need to have _some_ write permission for this?

pepijnve: The necessary access rights are added just underneath this line. I think we can actually start from an empty set in this method. The only thing that's missing is a check for the 'group owner'. If that's being written I think you also need `WRITE_OWNER`.

hierynomus: Isn't that on line 405?

pepijnve: Yep, I'll add a review comment there.

pepijnve: Yes that's correct. We need to add `|| securityInfo.contains(SecurityInformation. GROUP_SECURITY_INFORMATION)` there I think.

PR #279, opened 2018-01-15
Exposing the host NetBIOS name via ConnectionInfo

skhare-r7:
As the title says..

Perhaps you have some comments whether information like this belongs in ConnectionInfo or at the Session level, is it is technically negotiated per Session.

### DIFF
--- a/src/main/java/com/hierynomus/smbj/auth/NtlmAuthenticator.java
+++ b/src/main/java/com/hierynomus/smbj/auth/NtlmAuthenticator.java
@@ -88,7 +88,7 @@ public class NtlmAuthenticator implements Authenticator {
                 logger.debug("Received NTLM challenge from: {}", challenge.getTargetName());
 
                 response.setWindowsVersion(challenge.getVersion());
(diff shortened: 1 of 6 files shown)


src/main/java/com/hierynomus/smbj/auth/NtlmAuthenticator.java
```diff
@@ -88,6 +88,8 @@ public AuthenticateResponse authenticate(final AuthenticationContext context, fi
                 logger.debug("Received NTLM challenge from: {}", challenge.getTargetName());
 
                 response.setWindowsVersion(challenge.getVersion());
+                response.setNetBIOSName(String.valueOf(challenge.getAvPairObject(AvId.MsvAvNbComputerName)));
```
tosmun-r7: This will convert `null` to `"null"`. I'd suggest adding a helper such as `getAvPairString` which will return null instead.

skhare-r7:
This has been stuck because of travis build for 5 hours now. Is there a way we can restart the build process?

PR #301, opened 2018-03-02
Extra DFS resolver concurrent usage safety 

pepijnve:
Make everything that's immutable final
Use atomic get/set for mutable fields


hierynomus:
@pepijnve Can you check the test failures with this PR?

pepijnve:
Fixed

PR #303, opened 2018-03-14
DFS target list preparation

pepijnve:


mikaelfj:
Hi Pepijn,
We are currently having som problems when transferring files to a DFS share with two physical fileservers "behind" the share.
As I see it the problem is related to issue #254 and I tried the workaround suggested by andershermansen - however this is still a hardcoding of which server to use when resolving the path.
I can see that your pull-request is not yet merged to the master, but does it offer an alternative solution to issue #254 and a solution that has been tried in a production environment?
Best regards
Mikael

hierynomus (approved):

PR #307, opened 2018-03-30
Ensure users don't get bitten by the path separator

hierynomus:


hierynomus:
@pepijnve Something like this you mean?

pepijnve:
yep, beat me to it.

PR #318, opened 2018-04-24
Add minimal Android build on Travis

pepijnve:


hierynomus:
Thanks @pepijnve!

.travis.yml
```diff
@@ -10,14 +9,29 @@ cache:
...
   - pip install --user codecov
 
-after_success:
-- codecov
+matrix:
+  include:
+    - language: java
+      jdk: oraclejdk8
+      after_success:
+        - codecov
```
hierynomus: I think running codecov once should suffice actually :)

src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,8 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
```
hierynomus: Can we generate the AndroidManifest on demand in the build. It would eliminate hardcoded version numbers.

build-android.gradle
```diff
@@ -0,0 +1,54 @@
...
+
+repositories {
+  jcenter()
+  google()
+}
+
+group = 'com.hierynomus'
+
+dependencies {
+  implementation "org.slf4j:slf4j-api:1.7.25"
```
hierynomus: Let's extract the version numbers to `gradle.properties` so that they're not duplicated (and become impossible to maintain)

src/main/AndroidManifest.xml
pepijnve: It seems like the file has to exist, but I did figure out how to inject values into it. Injecting the package name was not possible.

.travis.yml
pepijnve: Fixed

build-android.gradle
pepijnve: Fixed

PR #319, opened 2018-04-26
Add maximalAccess to TreeConnect and change in treeConnectResponse

yin19941005:
Add maximalAccess to TreeConnect class. Change the getMaximalAccess() in SMB2TreeConnectResponse to Set<AccessMask> instead of long. Change the test case to use Set<AccessMask> to compare instead of using long.

This should allow the client to perform a local checking before actual sending out a request which will get "ACCESS_DENIED" from the server.


hierynomus:
Thanks!

PR #320, opened 2018-04-27
Bug Fix: Wrong name on creating SpnegoAuthenticator Factory by reflection

yin19941005:
Fix wrong name on creating SpnegoAuthenticator Factory on SmbConfig by reflection. The nested class name should be named as "outerClass$nestedClass" instead of "outerClass.nestedClass".

This code is just add on [here](https://github.com/hierynomus/smbj/commit/975b806164e1e6fbbba3d3e63ed30a036cc52958#diff-7b69131483bcf0d3043d16a5bfbf49b8R90). Without this fix, you will always get "ClassNotFoundException" for the SpnegoAuthenticator Factory.


hierynomus:
Thanks, good catch!

PR #329, opened 2018-05-07
Make createDefaultConfig return a usable config on Android

pepijnve:


hierynomus:
Doesn't work @pepijnve, see the build log for the android build:

```
:compileDebugJavaWithJavac/home/travis/build/hierynomus/smbj/src/main/java/com/hierynomus/smbj/SmbConfig.java:27: error: cannot find symbol
import com.hierynomus.smbj.auth.SpnegoAuthenticator;
                               ^
  symbol:   class SpnegoAuthenticator
  location: package com.hierynomus.smbj.auth
/home/travis/build/hierynomus/smbj/src/main/java/com/hierynomus/smbj/SmbConfig.java:112: error: package SpnegoAuthenticator does not exist
            authenticators.add(new SpnegoAuthenticator.Factory());
```

pepijnve:
Rats, forgot that I had excluded that file from the build.
I guess we'll have to restore the reflection stuff then.

pepijnve:
@hierynomus I can fix this in two ways:
1. Load `SpnegoAuthenticator.Factory` using reflection
2. Work around the issue in the build script by excluding the real `SpnegoAuthenticator` class and friends and instead including a fake, empty, implementation.

Do you have a preference for one or the other? I'm inclined to go for 2. to avoid polluting the code with bits that are only needed to satisfy the CI system.

hierynomus:
Hmmm... Instead of the empty implementation I would include one that throws UnsupportedOperationException or something to that effect. That would immediately signal something is wrongly packaged if that comes up

pepijnve:
UnsupportedOperationException is a better idea indeed. Last commit compiled correctly locally; let's see if Travis is happy as well.

pepijnve:
Nope, still more work to be done

hierynomus:
I think you might've missed something in the commit, the current 'fake' file is fully empty ;)

pepijnve:
Indeed it is. How the heck did that happen??

hierynomus:
It seems that you do need the exclude in the default sourceSet for this to work.

pepijnve:
I'll keep on tinkering with this; ignore the noise. Unfortunately I can't experiment easily in my own fork. Travis CI builds are broken there due to the sonarqube stuff.

hierynomus:
Yeah, noticed that sonar doesn't like that indeed. will separate that and
codecov out into a QA build in the matrix. THen the regular builds should
be green.

2018-05-07 13:09 GMT+02:00 Pepijn Van Eeckhoudt <notifications@github.com>:

hierynomus:
It can be done by making it a separate sourceSet and adding that one to the classpath.

```
sourceSets {
  android {
    java {
      srcDir file('src/android')
    }
  }
  main.java.compileClasspath += sourceSets.android.output
}
```

pepijnve:
No go unfortunately. I was trying to do something similar with a sub-project, but there's a circular dependency in the code. If you don't mind moving the JavaSE dependent code to its own source dir, then that's probably going to be the easiest. i.e., `src/main/java`, `src/main/javase` and `src/main/android` or something like that.

pepijnve:
Last commit is an attempt at what I suggested. Trying to see if that works or not.

hierynomus:
Hmm... I don't like this.

Can't we do the following:
- build using the regular JDK
- run a compatibility test against android

Or keep the reflection stuff and make this change very minimal.

pepijnve:
I had mixed feelings about it as well.
FWIW it does make it very clear which code is shared, Java SE specific and Android specific.

Compiling against the android sdk remains the safest way to check for android compatibility though imo.

Not sure how we could orchestrate android tests on Travis. You would have to have a well-known SMB server to connect to...

hierynomus:
Maybe we should even keep it simpler and stay with the specific `AndroidConfig`. This is something sshj also has. It's also easy to test and works in the current infrastructure.

pepijnve:
> stay with the specific AndroidConfig

That still won't compile against the Android SDK though which is the problem I'm trying to solve here. As long as SmbConfig has a static dependency on `SpnegoAuthenticator.Factory` you'll get compile errors. The path I've been taking is an attempt to get rid of the reflection stuff.

I personally prefer the automatic config option since that allows things like the basic example from the readme (`SMBClient client = new SMBClient();`) to just work. Only force users to configure things when they need to.

A hybrid is possible of course; keep the reflection bits to workaround the compile errors and keep the autoconfig logic to allow the library to work on Android without requiring extra code.

hierynomus:
I think that might be the simplest solution.. Let's just do that, smallest commit footprint :)

pepijnve:
Take 3

hierynomus:
Let's see what Travis says, but I think you forgot to remove the import of Spnego ;)

pepijnve:
aaargh. that's what you get for juggling too many balls.

Issue #339, opened 2018-06-12
IndexOutOfBoundsException when resolving DFS referral response


dwissk:
Version: 0.7.0 from maven central

When [accessing the the first referral entry of a DFS root referral response](https://github.com/hierynomus/smbj/blob/master/src/main/java/com/hierynomus/msdfsc/ReferralCache.java#L114) an `IndexOutOfBoundsException` is thrown.

Stacktrace:
`java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.util.ArrayList.rangeCheck(Unknown Source)
	at java.util.ArrayList.get(Unknown Source)
	at com.hierynomus.msdfsc.ReferralCache$ReferralCacheEntry.<init>(ReferralCache.java:114)
	at com.hierynomus.smbj.paths.DFSPathResolver.handleRootOrLinkReferralResponse(DFSPathResolver.java:434)
	at com.hierynomus.smbj.paths.DFSPathResolver.handleReferralResponse(DFSPathResolver.java:421)
	at com.hierynomus.smbj.paths.DFSPathResolver.getReferral(DFSPathResolver.java:402)
	at com.hierynomus.smbj.paths.DFSPathResolver.sendDfsReferralRequest(DFSPathResolver.java:390)
	at com.hierynomus.smbj.paths.DFSPathResolver.step6(DFSPathResolver.java:225)
	at com.hierynomus.smbj.paths.DFSPathResolver.step5(DFSPathResolver.java:186)
	at com.hierynomus.smbj.paths.DFSPathResolver.step2(DFSPathResolver.java:124)
	at com.hierynomus.smbj.paths.DFSPathResolver.step1(DFSPathResolver.java:107)
	at com.hierynomus.smbj.paths.DFSPathResolver.resolve(DFSPathResolver.java:94)
	at com.hierynomus.smbj.paths.DFSPathResolver.resolve(DFSPathResolver.java:75)
	at com.hierynomus.smbj.share.DiskShare.createFileAndResolve(DiskShare.java:82)
	at com.hierynomus.smbj.share.DiskShare.createFileAndResolve(DiskShare.java:92)
	at com.hierynomus.smbj.share.DiskShare.open(DiskShare.java:70)
	at com.hierynomus.smbj.share.DiskShare.exists(DiskShare.java:173)
	at com.hierynomus.smbj.share.DiskShare.folderExists(DiskShare.java:169)
`

[...]

PR #340, opened 2018-06-13
throw PathResolveException for deactivated referral links

dwissk:
so that deactivated referral links are skipped when listing folder contents

#339


dwissk:
closing as the code changes have found their way into master already

PR #354, opened 2018-07-10
Implement Async Create Request/Response with Oplock implementation.

yin19941005:
Add Oplock related messages and enum classes. Add oplock related NtStatus. Read messageId in messageConverter to determine oplock message type. Implement equals and hashCode for SMB2FileId to let user easier to compare. Implement the Async Create Request/Response and Oplock Break Notification by implementing notificationHandler with SMBEvent. Capturing the messageId of async create before send and capturing the Oplock Break Notification and Create Response on receiving in Connection and delivery them to user with ensuring the sequence is exactly the same as TCP layer. Add more create API to achieve the condition mentioned before. Public some methods to allow the user to create the diskEntry by his/her own.

Add taskQueue and test case for Async Create and oplock related implementation.


yin19941005:
Hi @hierynomus,

The `taskQueue` class is copied from the [Vert.x](https://github.com/eclipse/vert.x/blob/master/src/main/java/io/vertx/core/impl/TaskQueue.java). Base on my understanding to `Apache License 2.0`, as I keep the license and add comment for each modification, it will be fine to use it. Just special remark for this sensitive licensing issue.

yin19941005:
Hi @pepijnve, 

Indeed the SingleThreadExecutor can provide the similar effect. If you are considering there are only one DiskShare instance exists, it make no difference. The difference between `SingleThreadExecutor` and `TaskQueue` is performance optimization. The goal is guarantee sequence for notification within a DiskShare, i.e. notifying DiskShare A and notifying DiskShare B at the same time is allowed. To achieve this goal, I come up with 3 approaches : 

1. Use single thread pool for each DiskShare
2. Use a single thread pool for whole Smbj
3. Use a task queue per DiskShare and having multi-threaded thread pool for whole Smbj.

<h4>Use single thread pool for each DiskShare</h4>
Easy to implement and easy to guarantee the event sequence. But creating thread is an expensive operation operation, which should avoid. Also, having dedicated thread for each DiskShare is obviously over killed. The context switching overhead is high as well.

<h4>Use a single thread pool for whole Smbj</h4>
Easy to implement and easy to guarantee the event sequence. However, using single thread may become performance bottomneck as the asynchronous event scale up when numbers of DiskShare increase. Also, I had planned to add more and more asynchronous event in smbj (e.g. Notify, Lease), using single thread for whole Smbj may become performance bottomneck.

<h4>Use a task queue per DiskShare and having multi-threaded thread pool for whole Smbj</h4>
The user is flexible to determine how many threads they needed. Multi-threading across different DiskShare instance is allowed.

<h4>Conclusion</h4>
As a result, I think using a task queue per DiskShare and having multi-threaded thread pool for whole Smbj is the best choice.

src/main/java/com/hierynomus/protocol/commons/concurrent/TaskQueue.java
```diff
@@ -0,0 +1,120 @@
+/*
```
hierynomus: Let's not copy-paste this. The license is not correct now anymore (wrong attribution).
Why aren't we using a simple `LinkedBlockingQueue` to queue the events?

src/main/java/com/hierynomus/smbj/share/DiskShare.java
```diff
@@ -434,17 +565,115 @@ public void setSecurityInfo(SMB2FileId fileId, Set<SecurityInformation> security
...
+     * Handler for handing the oplock break notification event from server. 3.2.5.19 Receiving an SMB2 OPLOCK_BREAK Notification.
+     *
+     * @param oplockBreakNotification received oplock break notification from server.
+     */
+    @Handler
+    @SuppressWarnings("unused")
+    private void oplockBreakNotification(final OplockBreakNotification oplockBreakNotification) {
+        try {
+
+            final SMB2FileId fileId = oplockBreakNotification.getFileId();
```
hierynomus: Shouldn't we check here whether the `OplockBreakNotification.fileId` is actually handled by this `DiskShare`?

src/main/java/com/hierynomus/smbj/share/DiskShare.java
```diff
@@ -434,17 +565,115 @@ public void setSecurityInfo(SMB2FileId fileId, Set<SecurityInformation> security
...
+                // Preventing the improper use of handler (holding the thread). if holding thread, timeout exception will be throw.
+                taskQueue.execute(new Runnable() {
+                    @Override
+                    public void run() {
+                        notificationHandler.handleOplockBreakNotification(oplockBreakNotification);
+                    }
+                }, notifyExecutor);
+            }else {
+                logger.warn("FileId {}, NotificationHandler not exist to handle Oplock Break.", fileId);
+                throw new IllegalStateException("NotificationHandler not exist to handle Oplock Break.");
```
hierynomus: Why is this an exception, and the `createRequestNotification` does not throw an exception?

src/main/java/com/hierynomus/mssmb2/SMB2OplockBreakLevel.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+package com.hierynomus.mssmb2;
+
+import com.hierynomus.protocol.commons.EnumWithValue;
+
+/**
+ * [MS-SMB2].pdf 2.2.23 SMB2 OPLOCK_BREAK Notification - OplockLevel
+ */
+public enum SMB2OplockBreakLevel implements EnumWithValue<SMB2OplockBreakLevel> {
```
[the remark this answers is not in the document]
yin19941005: This is because the valid set is different between requesting oplock and breaking oplock. Only `LEVEL_NONE` and `LEVEL_II` are **valid for breaking oplock**. Base on my understanding, each enum class should only contain the valid values (and also null in Java). Adding `valid check` functions maybe is an alternative but `valid check` function check on **run time**. Separating this enum will force the checking on **compile time**. In my opinion, error checking on compile time is better than on run time. For me, I think simply separate this enum should make the code looks simpler than calling `valid check` function.

src/main/java/com/hierynomus/mssmb2/messages/SMB2OplockBreak.java
```diff
@@ -0,0 +1,49 @@
...
+public abstract class SMB2OplockBreak extends SMB2Packet {
+
+    protected SMB2OplockBreakLevel oplockLevel;
+    protected SMB2FileId fileId;
+
+    protected SMB2OplockBreak() {
+        super();
+    }
+
+    protected SMB2OplockBreak(int structureSize, SMB2Dialect dialect, SMB2MessageCommandCode messageType, long sessionId) {
```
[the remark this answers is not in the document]
yin19941005: Indeed, will change it.

src/main/java/com/hierynomus/mssmb2/messages/SMB2OplockBreakFactory.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public SMB2OplockBreak read(SMBBuffer buffer) throws Buffer.BufferException {
+        // 3.2.5.19 Receiving an SMB2 OPLOCK_BREAK Notification
+        // If the MessageId field of the SMB2 header of the response is 0xFFFFFFFFFFFFFFFF,
+        // this MUST be
+        // processed as an oplock break indication.
+        buffer.skip(24);
+        byte[] messageId = buffer.readRawBytes(8);
+        buffer.rpos(0);
+        final boolean isBreakNotification = Arrays.equals(messageId, (new byte[]{(byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF}));
```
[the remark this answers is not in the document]
yin19941005: Will add `private static final long breakMessageId = -1;` on SMB2OplockBreakFactory and check by `final boolean isBreakNotification = messageId == breakMessageId;` instead of using byte array.

src/main/java/com/hierynomus/mssmb2/messages/SMB2OplockBreakFactory.java
```diff
@@ -0,0 +1,52 @@
...
+
+        // TODO: Use structureSize as well to determine oplock and lease.
+        // buffer.skip(64);
+        // final int structureSize = buffer.readUInt16();
+        // buffer.rpos(0);
+
+        if(isBreakNotification) {
```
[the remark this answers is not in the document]
yin19941005: I guess no spacing issue on this line?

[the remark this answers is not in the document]
hierynomus: Look at the other places, a space between `if` and `(` is missing... (Here and in other places).

src/main/java/com/hierynomus/mssmb2/messages/SMB2OplockBreakFactory.java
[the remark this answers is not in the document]
hierynomus: 👍

src/main/java/com/hierynomus/mssmb2/SMB2OplockBreakLevel.java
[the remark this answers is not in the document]
hierynomus: Please model it according to spec, there is no `SMB2_OPLOCK_BREAK_LEVEL`, only `SMB2_OPLOCK_LEVEL`. With different enums we need to have translation code in place to transform one into the other. Please merge them.

src/main/java/com/hierynomus/protocol/commons/concurrent/FutureWrapper.java
```diff
@@ -0,0 +1,58 @@
...
+ * limitations under the License.
+ */
+package com.hierynomus.protocol.commons.concurrent;
+
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+public class FutureWrapper<T extends U, U> implements Future<T> {
```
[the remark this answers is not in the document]
yin19941005: Currently, FutureWrapper is used to deal with issue on [Connection publishing `AsyncCreateResponseNotification`](https://github.com/hierynomus/smbj/pull/354/files/9df5b0f34ef7143aac8f9958d5d727bf2337040e#diff-3d85ea955c455e6d3fd678f9f7e6e7d4R459). In Java, `Dogs` is a child of `Animal` but `List<Dogs>` is not a child of `List<Animal>` as well as `Future<Dogs>` is not a child of `Future<Animal>`.

However, I just read the details for
```
// [MS-SMB2].pdf 3.2.5.1.8 Processing the Response
        outstandingRequests.receivedResponseFor(messageId).getPromise().deliver(packet);
```
and just finds that it only delivering the packet to Future and do nothings more. 

I was thinking there are some process is required to handle for `3.2.5.1.8 Processing the Response` so I just kept it in `Future` format and get the `Future` just as the synchronize create.

I think I should simply publish the `SMB2CreateResponse` on the Connection and remove this `FutureWrapper` class.

src/main/java/com/hierynomus/smbj/share/DiskShare.java
yin19941005: Because the `createRequestNotification` and `createResponseNotification` is always called even you are using synchronize `create` (the existing `create`). The is a valid use case for a user to using the synchronize `create` and not set the NotificationHandler (as this is intended to handle async event only).

However, it is not a valid use case for using any level of oplock higher than `None` on any kind of `create` command. Because the `OplockBreakNotification` is always "Async" (or I should say Server initiated?).

src/main/java/com/hierynomus/smbj/share/DiskShare.java
yin19941005: I don't think we need to check for fileId. The client **MUST** ignore it if `fileId` not found.

Base on document,
>3.2.5.1.2 Finding the Application Request for This Response
>The client MUST locate the request for which this response was sent in reply by locating the request in
Connection.OutstandingRequests using the MessageId field of the SMB2 header. If the request is
not found, the response MUST be discarded as invalid.
If the MessageId is 0xFFFFFFFFFFFFFFFF, this is not a reply to a previous request, and the client
MUST NOT attempt to locate the request, but instead process it as follows:
If the command field in the SMB2 header is SMB2 OPLOCK_BREAK, it MUST be processed as specified
in 3.2.5.19. Otherwise, the response MUST be discarded as invalid.

>3.2.5.19.1 Receiving an Oplock Break Notification
>The client MUST locate the open in the Session.OpenTable using the FileId in the Oplock Break
Notification following the SMB2 header. If the open is **not found**, the oplock break indication **MUST be
discarded**, and **no further processing is required**.

And currently, it seems the `DiskShare` didn't record for the openedHandle. I don't think we should add it just because an unnecessary checking.

But indeed reminds me something. The `fileId` is only unique within a `session`. We should check the `oplockBreakNotification.getHeader().getSessionId()` match `DiskShare.session.getSessionId()` or not before notifying the user.

src/main/java/com/hierynomus/protocol/commons/concurrent/TaskQueue.java
yin19941005: I see your point. Let me check how to implement a "TaskQueue" with a `LinkedBlockingQueue`.

src/main/java/com/hierynomus/smbj/share/DiskShare.java
hierynomus: At this moment multiple `DiskShare`s will be receiving the message from the bus. You need to check which one needs to process the message, so you _do_ need to check both the `sessionId` as well as the `fileId`.

yin19941005: Just to be sure, is it sufficient to confirm the `DiskShare` is same by checking the `oplockBreakNotification.getHeader().getTreeId()` match `DiskShare.treeConnect.getTreeId()`?

hierynomus: Ok, that makes that we need to use a contextual bus per connection. Because it could very well be that the same fileId exists on multiple different connections (to different hosts).

yin19941005:
Hi @hierynomus,

The commit I just pushed should fixed some issue. The remaining outstanding issue : 

TODO List:

1. Merge `SMB2_OPLOCK_LEVEL` and `SMB2_OPLOCK_BREAK_LEVEL`
2. Remove the SMB2MessageCommandCode from the signature, it is always SMB2_OPLOCK_BREAK for this subtree.
3. Code format fix
4. Implement `TaskQueue` with a `LinkedBlockingQueue`
5. Add a `Connection` internal EventBus for each `Connection`. Use the internal EventBus instead of the global EventBus for every `Notification`.
6. Make `AbstractAsyncNotification` is used by Async Request only. (To recording `messageId` and used for should handle when receive in Async Response)
7. Implement a internal centralize messageId callback to record all implemented notificationHandler MessageId?

Please let me know if the list missed some issue you had mentioned before.

For issue 6 and 7, I guess while I complete issue 7 and `AbstractAsyncNotification` will just disappear and issue 6 no longer exist. But I keep it on list for reference.

yin19941005:
Hi @hierynomus,

I guess I had already change the code to fit all your change request. 

For `TaskQueue`, I had changed it as interface (default using single thread executor) and allow user to pass their own `TaskQueue` in to use in config. I tried to implement it but I didn't sure I had implemented it correctly so I change it as interface for more flexibility. Is this okay for you?

hierynomus:
Can you take a look at the codacy messages? There are a few degradations in quality.

yin19941005:
No problem. Will try to fix the codacy stuffs.

yin19941005:
Hi @hierynomus,

The codacy issue had been fixed. However I still got error:

> security/snyk - build.gradle (hierynomus (GitHub marketplace)) — Failed to detect issues

I guess this is caused by this branch is out of date?

PR #294, opened 2018-02-16
Copy issue 175

JozefDropco:
Hi, 
I tried to implement server-side copy of file. I dont know where you would like to place those files as it creates new concept into your code (looping and reacting on response within your code). I think there is a possibility to add progress listener.

PR #366, opened 2018-07-20
Remote copy take 2

pepijnve:
This is a rebased version of PR #294. Adds remote file copy support.

### DIFF
--- a/src/it/groovy/com/hierynomus/smbj/SMB2FileIntegrationTest.groovy
+++ b/src/it/groovy/com/hierynomus/smbj/SMB2FileIntegrationTest.groovy
@@ -17,6 +17,7 @@ package com.hierynomus.smbj
 
 import com.hierynomus.msdtyp.AccessMask
 import com.hierynomus.mserref.NtStatus
+import com.hierynomus.msfscc.fileinformation.FileStandardInformation
 import com.hierynomus.mssmb2.SMB2CreateDisposition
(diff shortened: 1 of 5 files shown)


pepijnve:
@hierynomus I've further simplified and reduced the PR initially created by @JozefDropco down to the absolute minimum. Code is now pretty straightforward I think.

hierynomus:
Code looks way better already indeed.

Two things:
1. Codacy isn't happy (haven't checked why)
2. The IntegrationTest will not run out of the box, as it does not upload the files to the host before testing.

pepijnve:
Codacy warning fixed.
Integration test now generates the source file itself.

pepijnve:
Note that in the old version I had added a special case for some server that responded in a non-standard way. That's what the
> Handle STATUS_INVALID_PARAMETER without a SRV_COPYCHUNK_RESPONSE
commit was about.

I can't remember which server that was or when this happened exactly, so I've omitted the code related to that for the time being.

felipecaparelli:
Thank you so much @hierynomus and @pepijnve! This feature provides great performance for softwares using SMB protocol.

felipecaparelli:
@hierynomus Do you will update the project version in the maven repository after this release?

hierynomus:
@felipecaparelli Version 0.8.0 has been pushed out to Maven central.

PR #371, opened 2018-08-10
Fix kerberos authentication against a linux host

jonasrutishauser:
This version works against Linux and Windows Hosts in our company.


hierynomus:
Thanks!

PR #373, opened 2018-08-13
Add IBM VM alternative

jonasrutishauser:
If the library is used in a IBM JVM the `com.sun.security.jgss` classes are placed in the `com.ibm.security.jgss` package.
The IBM variant is called via reflection.


jonasrutishauser:
Improved implementation.

If the minimal version would be java 8 i would have used `java.lang.invoke.LambdaMetafactory`.

PR #388, opened 2018-09-19
append to file option added

JozefDropco:
Hi hierynomus, 
I find useful to add the option of appending to existing file. We have a usecase where the file needs to be appended instead of overwriting. These small changes allows it


hierynomus:
Thanks @JozefDropco. Could you add some (integration) tests for this? Then (except for some reformatting) it is ready to be merged.

JozefDropco:
Hi @hierynomus ,
I added basic integration test.
Kind regards,
jozef

PR #394, opened 2018-10-02
Remove dependency on BouncyCastle ASN.1

hierynomus:
Can be merged once asn-one has been released.

### DIFF
--- a/build-android.gradle
+++ b/build-android.gradle
@@ -22,6 +22,7 @@ dependencies {
   implementation SLF4J_API
   implementation BCPROV_JDK15ON
   implementation MBASSADOR
+  implementation ASN_ONE
   testImplementation GROOVY_ALL
   testRuntimeOnly CGLIB_NODEP
   testRuntimeOnly OBJENESIS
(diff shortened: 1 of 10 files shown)


hierynomus:
TODO left:
[] Make BC optional
[] Ensure no BC references in anything but SecurityProvider package.

krivanekm-javlin:
I've tested it and it looks like BC is no longer required and all our tests have passed. However, I don't think we use neither Spnego nor NTLM authentication in our test environment.

If I understand it correctly, BC is now only required if I explicitly call `SmbConfig.Builder.withSecurityProvider(new BCSecurityProvider())`, it that correct?

hierynomus:
Yes, correct

krivanekm-javlin:
Just a minor comment, I've noticed that asn-one uses "com.hierynomus.asn-one" as the group ID, but smbj and sshj use "com.hierynomus". Is that intentional?

hierynomus:
> Just a minor comment, I've noticed that asn-one uses "com.hierynomus.asn-one" as the group ID, but smbj and sshj use "com.hierynomus". Is that intentional?

Totally unintentional! Well spotted. I'll move asn-one a level up.

PR #403, opened 2018-11-09
Move reading position to the end of message

gakisou:
Try to fix the issue 402


hierynomus:
This fix is a bit crude as in a compounded response it will break (i.e. the wpos is then a few packets ahead). I've fixed it better in #404

PR #447, opened 2019-03-30
Changes over time to support EMC/VNX and Isilon Devices

sbellan:
- Pass a non null workstation in NtlmAuthenticator to satisfy EMC/VNX
- closeSilently to call closeNoWait, where we dont want to wait long if the connection is bad.
- directory.list to take custom access, so that it can enumerate folder hierarchies with restricted permissions.
- Isilon sends a custom hintName containing the hostname of the server against the SMB2 spec. Ignore it.


src/main/java/com/hierynomus/smbj/share/Share.java
```diff
@@ -178,6 +178,11 @@ void closeFileId(SMB2FileId fileId) throws SMBApiException {
         sendReceive(closeReq, "Close", fileId, SUCCESS_OR_CLOSED, transactTimeout);
     }
 
+    void closeFileIdNoWait(SMB2FileId fileId) throws SMBApiException {
+        SMB2Close closeReq = new SMB2Close(dialect, sessionId, treeId, fileId);
+        sendReceive(closeReq, "Close", fileId, SUCCESS_OR_CLOSED, 0);
```
hierynomus: Passing `0` to the timeout means that it will wait potentially indefinitely (it won't timeout). The method name is very misleading. Why is this needed?

src/main/java/com/hierynomus/spnego/NegTokenInit.java
```diff
@@ -147,7 +152,7 @@ protected void parseTagged(ASN1TaggedObject asn1TaggedObject) throws SpnegoExcep
                 // Ignore mechListMIC for now...
                 break;
             default:
-                throw new SpnegoException("Unknown Object Tag " + asn1TaggedObject.getTagNo() + " encountered.");
+                logger.warn("Unknown Object Tag {}/{} encountered.. Ignoring", asn1TaggedObject.getTagNo(), asn1TaggedObject.getObject());
```
hierynomus: Why is this? The spec is pretty clear as to what this message should and should not contain. Also see the class level comment.

src/main/java/com/hierynomus/smbj/auth/NtlmAuthenticator.java
```diff
@@ -121,8 +121,8 @@ public AuthenticateResponse authenticate(final AuthenticationContext context, fi
                 Object msvAvFlags = challenge.getAvPairObject(AvId.MsvAvFlags);
                 if (msvAvFlags instanceof Long && ((long) msvAvFlags & 0x00000002) > 0) {
                     // MIC should be calculated
-                    NtlmAuthenticate resp = new NtlmAuthenticate(new byte[0], ntlmv2Response,
-                        context.getUsername(), context.getDomain(), null, sessionkey, EnumWithValue.EnumUtils.toLong(negotiateFlags),
+                    NtlmAuthenticate resp = new NtlmAuthenticate(new byte[24], ntlmv2Response,
```
hierynomus: Why are we passing a 24 byte array in?

src/main/java/com/hierynomus/smbj/share/Share.java
sbellan: Correct, the intent was to just do send instead of sendReceive. The reasoning is that when the routers/OS close a long running connection we get a IOException and in the finally block to we close the handle. In those cases we don't care about the responses nor we want to wait for the timeout period.

src/main/java/com/hierynomus/spnego/NegTokenInit.java
sbellan: Please see the attached trace, the isilon device sends the hostname in the hintName, whereas the spec says it should only contain, "not_defined_in_RFC4178@please_ignore".

image

src/main/java/com/hierynomus/smbj/auth/NtlmAuthenticator.java
sbellan: ok will add a config for the workstation name. Regarding the byte[24], I don't recollect now, if this was really needed or not. But spend quite some time at customer env, where the authentication was failing. I can keep the byte[24] out of this PR

Just for FYI, We have been using this in production for a long time now, and has been verified to work great against NetApp, Dell/EMC, Isilon etc after some of the above fixes.

src/main/java/com/hierynomus/spnego/NegTokenInit.java
hierynomus: I've fixed this in a different way now, see the current master. The response token with the negHints is an NegTokenInit2 token, which I totally missed in the initial implementation. Any server initiated NegTokenInit is always a NegTokenInit2 according to MS-SPNG.

Please remove this change from the PR.

sbellan: Thanks, removed my change now and added a test.

hierynomus:
thanks! It's merged.

PR #462, opened 2019-05-09
Expose some attributes from responses.

sbellan:


hierynomus:
The offset time looks fine, just wondering where the pathconsumed is needed.

sbellan:
We use the smb2 referral messages without using the dfs path resolve for
some legacy reasons. We have requirement given a path, give back the final
resolved path. We need the path consumed in that logic.

PR #350, opened 2018-07-03
Implementation for SMB2 lock command

yin19941005:
Implementation for SMB2 lock command. Add validation for SMB2Lock fields value. Add error NTSTATUS for SMB2 lock command.

PR #466, opened 2019-05-15
Fix the invalid checking for UInt64 (causing it only supports 63 bits)

yin19941005:
Checking is negative or not for 8 bytes variable (long) is meaningless when putting to another unsigned 8 bytes variable. And this causing only supports up to 63 bits instead of 64 bits (8 bytes).

The write request is not able to write to offset > 63 bits without this fix.

@hierynomus, based on my understanding on JAVA 8 unsigned operations added to `Long` (https://docs.oracle.com/javase/8/docs/api/java/lang/Long.html), it will keep using the normal long primitive type (which is signed) and using those unsigned operations to deal with the unsigned variable. So, I think we should read/write the UInt64 as same as long and only consider they are unsigned when using them.

### DIFF
--- a/src/main/java/com/hierynomus/protocol/commons/buffer/Endian.java
+++ b/src/main/java/com/hierynomus/protocol/commons/buffer/Endian.java
@@ -108,19 +108,12 @@ public abstract class Endian {
 
         @Override
         public <T extends Buffer<T>> void writeUInt64(Buffer<T> buffer, long uint64) {
-            if (uint64 < 0) {
-                throw new IllegalArgumentException("Invalid uint64 value: " + uint64);
-            }
             writeLong(buffer, uint64);
         }
 
         @Override
         public <T extends Buffer<T>> long readUInt64(Buffer<T> buffer) throws Buffer.BufferException {
-            long uint64 = (readUInt32(buffer) << 32) + (readUInt32(buffer) & 0xFFFFFFFFL);
-            if (uint64 < 0) {
-                throw new Buffer.BufferException("Cannot handle values > " + Long.MAX_VALUE);
-            }
-            return uint64;
+            return readLong(buffer);
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


hierynomus:
Hi @yin19941005,
Sorry to say, I disagree. The current implementation is the best practice for uints, as they need to be treated as positive integers, we can only support 63 bits in either `long` or `Long` (`Long` also uses 2s-complement). If we want to have 64bits unsigned support, we need to add a `BigInteger readUint64AsBigInteger()` method. 

Please also add testcases that prove this is needed. Keep in mind that an offset of >63 bits means the file is `9,223,372,036,854,775,807` bytes ==  `8,388,608 TiB` == `8,192 PiB`

hierynomus:
Also there is a `readLong` method which reads the full 64bits, but converts it to a negative number if the leftmost bit is set, as is the case with 2s-complement.

yin19941005:
> If we want to have 64bits unsigned support, we need to add a BigInteger readUint64AsBigInteger() method.

Yes, I do agree with this.

> Please also add testcases that prove this is needed. Keep in mind that an offset of >63 bits means the file is 9,223,372,036,854,775,807 bytes == 8,388,608 TiB == 8,192 PiB

Yes, I understand that's really rare case. Although this is kind of violating the SMB2 specification in my perspective, I agree that the case is so rare and basically not happening.

Actually I am working on some Hyper-V VM stuffs. I was using WireShark to check with what commands is calling when Hyper-V creating VM and simulate it with SmbJ. The Hyper-V will call the `SMB2 LOCK Request` for lock length `0xffffffffffffffff`. I try to simulate it with #350 and got `Buffer.BufferException`. I guess Hyper-V is dealing with issue related to `FILE_ATTRIBUTE_SPARSE_FILE` for some VM image may used. For sparse file, the file size is basically virtual, so using 64 bits is kind of reasonable.



> Also there is a readLong method which reads the full 64bits, but converts it to a negative number if the leftmost bit is set, as is the case with 2s-complement.

Yes, I had continue my testing on using `readLong`. I am just not sure when to use `UInt64` or `long`

Anyway, I would recommend the `putUInt64` javadoc stating it is not truly supports 64 bits.

ecki:
I just wanted to comment that I dont think its a good idea to use biginteger for uint64 for performance and compatibility reasons. It is quite niormal to treat a Java long as unsigned (thats why Java added all those unsigned methods). I would stick with long and make it clear what it means to get a negative long.

Issue #415, opened 2018-12-13
NPE in DiskShare


cliviu:
Hi,  Got NPE with 0.9.1 : Caused by: java.lang.NullPointerException
	at com.hierynomus.smbj.share.DiskShare.getDiskEntry(DiskShare.java:112)
	at com.hierynomus.smbj.share.DiskShare.open(DiskShare.java:67)
	at com.hierynomus.smbj.share.DiskShare.exists(DiskShare.java:187)
	at com.hierynomus.smbj.share.DiskShare.folderExists(DiskShare.java:183) 

Last log lines before were : 
2018-12-12 16:47:59,443|DFSPathResolver               |INFO |DFS Share Net-... does not cover \\path1 resolve through DFS
2018-12-12 16:47:59,443|DFSPathResolver               |INFO |Starting DFS resolution for \\path1
2018-12-12 16:47:59,443|DFSPathResolver               |INFO |DFS resolved path1-> path2
   What can be the problem ? 
  Thanks !

cliviu:
Any thoughts about this ? What can cause the problem ?  The NPE occurs in the "if" line : 
    SMB2CreateResponse response = responseContext.resp;
        if (response.getFileAttributes().contains(FILE_ATTRIBUTE_DIRECTORY)) {

Either "response" or the "fileAttributes"  must be null but I'm not sure what kind of external configuration can cause this problem ...

cliviu:
I have another hint: the DFS Resolution returns the same URL when the NPE is thrown ....

Disablez:
This could help...
https://github.com/hierynomus/smbj/pull/474

throwinto2:
I seem to be getting the exact same exception at the exact same code lines. I have multiple DFS shares set up on various servers and it seems like only some encounter this problem. They are all in the same domain, same setup and all shares are set up in the same way.
I sadly cannot pinpoint the cause. At first it looked like the DFS shares on a domain controller were more stable, but this was seemingly fixed in 2017. What seemed to work to a certain extent is creating a new DFS namespace with new shares. Maybe the DFS namespace cache is the cause..
Did #474 and #342 resolve this problem? I'm hoping it will be available in the next version when it is released.

davisnw:
I found that at least in my case, the `NullPointerException` only occurs if I reuse `SmbClient`.  See also #531

Issue #342, opened 2018-06-18
Is DFSPathResolver thread safe?


jborza:
Hi,

I've been getting this exception accessing DFS shares

```
java.lang.IllegalStateException: Could not find referral cache entry for DFSPath{[server.domain.com, dfs]}
        at com.hierynomus.smbj.paths.DFSPathResolver.step9(DFSPathResolver.java:291)
	at com.hierynomus.smbj.paths.DFSPathResolver.step2(DFSPathResolver.java:127)
	at com.hierynomus.smbj.paths.DFSPathResolver.step1(DFSPathResolver.java:107)
	at com.hierynomus.smbj.paths.DFSPathResolver.resolve(DFSPathResolver.java:94)
	at com.hierynomus.smbj.paths.DFSPathResolver.resolve(DFSPathResolver.java:75)
	at com.github.jborza.camel.component.smbj.dfs.DfsResolver.resolve(DfsResolver.java:31)
```
not really easy to reproduce, but it usually occurs in two scenarios:
1. multiple threads were accessing the share at same time and this happened seemingly at random after some time
2. single thread was accessing the share many times (recursive listing of folders) and this eventually

According to the spec ( https://msdn.microsoft.com/en-us/library/cc227028.aspx ) step9 should occur only if step2 finds out that the item is expired. Step2 actually checks whether a `state.path` is expired

```
        ReferralCache.ReferralCacheEntry lookup = referralCache.lookup(state.path);
        ...
        if (lookup.isExpired()) { // Expired LINK target
            return step9(session, state, lookup); // Resolve Link Referral
        }
```

and step9 looks up just the root

```
        DFSPath rootPath = new DFSPath(state.path.getPathComponents().subList(0, 2));
        ReferralCache.ReferralCacheEntry rootReferralCacheEntry = referralCache.lookup(rootPath);
        if (rootReferralCacheEntry == null) {
            throw new IllegalStateException("Could not find referral cache entry for " + rootPath);
        }
```

Technically when the lookup exists in the cache, the root should exist as well, as lookup is something like server.domain.com/dfs/folder/subfolder

Question 1: Is smbj SMBClient intended to be used from a single thread? At least the DFS cache bits seem thread-safe due to the use of `volatile entry` and `AtomicReferenceFieldUpdater`. 
[...]

PR #474, opened 2019-06-04
Issue #342

Disablez:
Fixes a problem where the root DFS entry would not get resolved when some other entries were expired.
This code clears the expired cache node and resolves it again as if it was the first time.
Probably fixes issue #415 as well.


src/main/java/com/hierynomus/smbj/paths/DFSPathResolver.java
```diff
@@ -301,7 +301,8 @@ private DFSPath step9(Session session, ResolveState state, ReferralCache.Referra
         DFSPath rootPath = new DFSPath(state.path.getPathComponents().subList(0, 2));
         ReferralCache.ReferralCacheEntry rootReferralCacheEntry = referralCache.lookup(rootPath);
         if (rootReferralCacheEntry == null) {
-            throw new IllegalStateException("Could not find referral cache entry for " + rootPath);
```
nddipiazza, line 304: maybe log that it happened?

src/main/java/com/hierynomus/msdfsc/ReferralCache.java
```diff
@@ -87,6 +87,11 @@ public ReferralCacheEntry lookup(DFSPath dfsPath) {
         return referralEntry;
     }
 
+    public void clear(DFSPath dfsPath) {
+        List<String> pathComponents = dfsPath.getPathComponents();
+        cacheRoot.deleteExpiredReferralEntry(pathComponents.iterator());
```
nddipiazza, line 92: instead of sending iterator, wouldn't it be a little cleaner to just to send the `List`?

src/main/java/com/hierynomus/msdfsc/ReferralCache.java
```diff
@@ -219,6 +224,21 @@ ReferralCacheEntry getReferralEntry(Iterator<String> pathComponents) {
             return ENTRY_UPDATER.get(this);
         }
 
+        void deleteExpiredReferralEntry(Iterator<String> pathComponents) {
+            if (this.entry != null && this.entry.isExpired() &&
+                !this.entry.isRoot() /*&& !this.entry.isLink()*/) {
```
nddipiazza, line 229: remove the commented out code. add a comment if necessary.

src/main/java/com/hierynomus/smbj/paths/DFSPathResolver.java
Disablez: Added logging.

src/main/java/com/hierynomus/msdfsc/ReferralCache.java
Disablez: It's recursive and the other functions used an iterator; changed it to Lists anyway.

src/main/java/com/hierynomus/msdfsc/ReferralCache.java
Disablez: Done.

hierynomus (approved):

Issue #406, opened 2018-11-14
DFS Reuse Session


SpeckPgh:
When leveraging DFS it appears to be creating a new nestedSession for every call to connectShare, even if it already has a nested session for the resolved host.   Is it possible to change the library to leverage existing nestedsession, rather than creating a new one every time?  Either by default or as a parameter on the SMBConfig?

EX:

2018-11-15 11:42:13.250  INFO 65213 --- [ead #10 - Split] c.hierynomus.smbj.paths.DFSPathResolver  : DFS Share EBA does not cover \\SOMESERVER\EBA\RCSSupport\OCPF\OCPFTest\MAGREL\IMGEDFS3\108\93, resolve through\
 DFS
2018-11-15 11:42:13.251  INFO 65213 --- [ead #10 - Split] c.hierynomus.smbj.paths.DFSPathResolver  : Starting DFS resolution for \\SOMESERVER\EBA\RCSSupport\OCPF\OCPFTest\MAGREL\IMGEDFS3\108\93
2018-11-15 11:42:13.251  INFO 65213 --- [ead #10 - Split] c.hierynomus.smbj.paths.DFSPathResolver  : DFS resolved \\SOMESERVER\EBA\RCSSupport\OCPF\OCPFTest\MAGREL\IMGEDFS3\108\93 -> \\smb-prd.somewhere.somehow.c\
om\ocpf_test\MAGREL\IMGEDFS3\108\93
2018-11-15 11:42:13.280  INFO 65213 --- [ead #10 - Split] c.hierynomus.smbj.connection.Connection  : Successfully authenticated SOMEUSER on smb-prd.somewhere.somehow.com, session is 562027080913580721
2018-11-15 11:42:13.280  INFO 65213 --- [ead #10 - Split] com.hierynomus.smbj.session.Session      : Connecting to \\smb-prd.somewhere.somehow.com\ocpf_test on session 562027080913580721
2018-11-15 11:42:13.307  INFO 65213 --- [ead #10 - Split] com.upmce.demo.utils.Smb                 : GETTING SHARE DFS/share
2018-11-15 11:42:13.309  INFO 65213 --- [ead #10 - Split] c.hierynomus.smbj.paths.DFSPathResolver  : DFS Share EBA does not cover \\SOMESERVER\EBA\RCSSupport\OCPF\OCPFTest\MAGREL\IMGEDFS3\105\191, resolve throug\
h DFS
[...]

PR #489, opened 2019-07-31
Reuse nested session when hostname is same (#406)

andershermansen:
Session is intended to be reused, so also reuse nested sessions when host is same.

This has been tested on DFS where I previously got STATUS_INSUFFICIENT_RESOURCES but not with this change.

### DIFF
--- a/src/main/java/com/hierynomus/smbj/session/Session.java
+++ b/src/main/java/com/hierynomus/smbj/session/Session.java
@@ -40,7 +40,9 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.Future;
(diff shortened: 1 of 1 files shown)


mikhasd:
This PR would fix #496

src/main/java/com/hierynomus/smbj/session/Session.java
```diff
@@ -61,7 +63,7 @@
     private SMBEventBus bus;
     private final PathResolver pathResolver;
     private TreeConnectTable treeConnectTable = new TreeConnectTable();
-    private List<Session> nestedSessions = new ArrayList<>();
+    private Map<String, Session> nestedSessions = new HashMap<>();
```
hierynomus: The change seems valid, however now there is potential race condition because there might be two threads trying to create the same nested session. One of these will be "lost". Please fix this case.

hierynomus:
Let's continue this PR in #499

PR #490, opened 2019-08-01
Hotfix read request always read credit maximum may conflict the byte range exclusive lock

yin19941005:
Hotfix read request always read credit maximum may conflict the byte range exclusive lock

Use the getPayloadSize() method instead of simply calculate from the creditsAssigned to avoid reading extra bytes and fix the conflict byte range exclusive lock issue.

Hi @hierynomus, SMB2 protocol is supporting byte range lock (both share and exclusive), you may check [[MS-SMB2] 2.2.26 SMB2 LOCK Request](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/6178b960-48b6-4999-b589-669f88e9017d) for details. Sending read request reading more than user required may cause conflict with the exclusive byte range lock and resulting a failed response (which expected to success). I think it should always read the exact bytes needed.


hierynomus:
The change seems valid, but there is no testcase to prevent regression (or prove the fix). Is it possible to add one?

yin19941005:
Ok, I see your point. Let me add some test cases to prevent bug regression.

PR #499, opened 2019-10-04
reuse nested sessions by host -- fixes #498 #406

### LINKED ISSUE #498: DFS resolution redundancy
I have a performance issue. Likely it's a feature request. I'm using SMBJ to copy a bunch of files to a DFS enabled share. 

I have a loop like this:

```java
for (Path sourcePath: sourcePaths) {
    final targetPath = BACKSLASH_JOINER.join(targetRoot, sourcePath.getFileName());
    try (com.hierynomus.smbj.share.File file = share.openFile(targetPath,
            EnumSet.of(AccessMask.GENERIC_ALL), null, SMB2ShareAccess.ALL, FILE_CREATE, null);
            OutputStream shareOutStream = file.getOutputStream(false)) {
        Files.copy(sourcePath, shareOutStream);
    }
}
```

With this code DFS resolution is performed by SMBJ for each file, but it always resolves to the same host and share in this case as I'm writing to the same target folder, so it's completely redundant.

I'd like to avoid these steps:

+ an extra DFS resolution;
+ connection / auth / etc to the same server for each file.

These are big performance hit especially when the remote host is slow or there is a network latency as in my case the: share is on a different continent.

I'd expect that somehow I can open a `Directory` for writing, and re-use it within the loop instead of specifying a string path which gets resolved each time. Or perhaps an API call where I could provide a pre-resolved path.

zsoltm:
Based on solution by @andershermansen at #489 but with thread safety.


goroskob:
Any chance this PR could be merged?

hierynomus (dismissed):

hierynomus (approved):

mkueng1:
I got stuck on this too. Is there already a planned date for the next release?

cliviu:
Yes, me too. Any planned date for the release ?

PR #545, opened 2020-04-03
issue-544 fix (implement DiskShare.mkdirs procedure which could creat…

AndreyFomichev:
issue-544 fix (implement DiskShare.mkdirs procedure which could create several nested directories in one call, for example, 2345 \ 3456 \ 4453 \ 123123.txt)

### DIFF
--- a/src/main/java/com/hierynomus/smbj/share/DiskShare.java
+++ b/src/main/java/com/hierynomus/smbj/share/DiskShare.java
@@ -261,35 +261,6 @@ public class DiskShare extends Share {
     }
 
     /**
-     * Create a set of nested sub-directories in the given path, for example, 2345 \ 3456 \ 4453 \ 123123.txt
-     */
-    public void mkdirs(String path) throws SMBApiException {
-        String currPath = path;
-        StringBuilder currDir = new StringBuilder();
-        while (path.startsWith("\\"))
-            currPath = currPath.substring(1, currPath.length() - 1 );
-
-        if (currPath.indexOf( "\\") <= 0 && !folderExists(currPath)) {
-            mkdir(currPath);
-        }   else {
-            for (String dir: currPath.split(String.valueOf("\\\\"))) {
-                currDir = currDir.append(dir).append("\\");
-                if (currDir != null && currDir.length() > 0 && !folderExists(currDir.toString())) {
-                    Directory fileHandle = openDirectory(currDir.toString(),
-                        of(FILE_LIST_DIRECTORY, FILE_ADD_SUBDIRECTORY),
-                        of(FILE_ATTRIBUTE_DIRECTORY),
-                        ALL,
-                        FILE_CREATE,
-                        of(FILE_DIRECTORY_FILE)
-                    );
-                    fileHandle.close();
-                    //System.out.println("Dir " + currDir + " created.");
-                }
-            }
-        }
-    }
-
-    /**
      * Get information about the given path.
      **/
     public FileAllInformation getFileInformation(String path) throws SMBApiException {
--- a/src/main/java/com/hierynomus/smbj/share/DiskShare.java
+++ b/src/main/java/com/hierynomus/smbj/share/DiskShare.java
@@ -261,6 +261,35 @@ public class DiskShare extends Share {
     }
 
     /**
+     * Create a set of nested sub-directories in the given path, for example, 2345 \ 3456 \ 4453 \ 123123.txt
+     */
+    public void mkdirs(String path) throws SMBApiException {
+        String currPath = path;
+        StringBuilder currDir = new StringBuilder();
+        while (path.startsWith("\\"))
+            currPath = currPath.substring(1, currPath.length() - 1 );
+
+        if (currPath.indexOf( "\\") <= 0 && !folderExists(currPath)) {
+            mkdir(currPath);
+        }   else {
+            for (String dir: currPath.split(String.valueOf("\\\\"))) {
+                currDir = currDir.append(dir).append("\\");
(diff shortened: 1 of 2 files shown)


src/main/java/com/hierynomus/smbj/share/DiskShare.java
```diff
@@ -260,6 +260,35 @@ public void mkdir(String path) throws SMBApiException {
         fileHandle.close();
     }
 
+    /**
+     * Create a set of nested sub-directories in the given path, for example, 2345 \ 3456 \ 4453 \ 123123.txt
+     */
+    public void mkdirs(String path) throws SMBApiException {
```
hierynomus, line 266: Move this to the `SmbFiles` utility class. The `mkdirs` is not a part of the protocol directly, and now has too much path handling magic. Also do not commit commented out code.

hierynomus:
Also please add tests...

src/main/java/com/hierynomus/smbj/share/DiskShare.java
AndreyFomichev: Mkdirs procedure has been moved to the SmbFiles utility class, comments are removed.

AndreyFomichev:
> 
> 
> Also please add tests...

i do not know Groovy, which is used to tests here. Can you help me by writing these tests ?

src/main/java/com/hierynomus/smbj/utils/SmbFiles.java
```diff
@@ -64,4 +74,31 @@ public static int copy(File source, DiskShare share, String destPath, boolean ov
...
     }
 
+    /**
+     * Create a set of nested sub-directories in the given path, for example, 2345 \ 3456 \ 4453 \ 123123.txt
+     */
+    public void mkdirs(DiskShare diskShare, String path) throws SMBApiException {
+        String currPath = path;
+        StringBuilder currDir = new StringBuilder();
+        while (path.startsWith("\\"))
+            currPath = currPath.substring(1, currPath.length() - 1 );
```
hierynomus, line 84: Use `substring(int beginIndex)`, now you also chop off a character at the end of the string. This is why tests are important.

src/main/java/com/hierynomus/smbj/utils/SmbFiles.java
```diff
@@ -64,4 +74,31 @@ public static int copy(File source, DiskShare share, String destPath, boolean ov
...
+
+        if (currPath.indexOf( "\\") <= 0 && !diskShare.folderExists(currPath)) {
+            diskShare.mkdir(currPath);
+        }   else {
+            for (String dir: currPath.split(String.valueOf("\\\\"))) {
+                currDir = currDir.append(dir).append("\\");
+                if (currDir != null && currDir.length() > 0 && !diskShare.folderExists(currDir.toString())) {
```
hierynomus, line 91: Pretty sure that `currDir` is always `!= null` as it is initialized at the start of the method.

hierynomus: Also it's length is also `>0` as you've appended just above this line.

src/main/java/com/hierynomus/smbj/utils/SmbFiles.java
```diff
@@ -64,4 +74,31 @@ public static int copy(File source, DiskShare share, String destPath, boolean ov
...
+                    Directory fileHandle = diskShare.openDirectory(currDir.toString(),
```
hierynomus, line 92: Why not directly call `diskShare.mkdir` here instead of duplicating that code?

src/main/java/com/hierynomus/smbj/utils/SmbFiles.java
```diff
@@ -64,4 +74,31 @@ public static int copy(File source, DiskShare share, String destPath, boolean ov
         return r;
     }
 
+    /**
+     * Create a set of nested sub-directories in the given path, for example, 2345 \ 3456 \ 4453 \ 123123.txt
+     */
+    public void mkdirs(DiskShare diskShare, String path) throws SMBApiException {
```
hierynomus, line 80: You can simplify this greatly by inverting the logic. Instead of starting from the beginning, find the longest path segment that exists and only create the tail end.

For deeply nested paths for which a big part already exists, this will save a lot of remote calls.

src/main/java/com/hierynomus/smbj/utils/SmbFiles.java
AndreyFomichev: fixed

src/main/java/com/hierynomus/smbj/utils/SmbFiles.java
AndreyFomichev: fixed

src/main/java/com/hierynomus/smbj/utils/SmbFiles.java
AndreyFomichev: fixed

src/main/java/com/hierynomus/smbj/utils/SmbFiles.java
AndreyFomichev: I'm sorry, but I can’t determine how to implement this. Can you help me to fix this issue ?

hierynomus:
Simplified and merged.

src/main/java/com/hierynomus/smbj/utils/SmbFiles.java
vitr1988: why not use static method for mkdirs?

PR #589, opened 2020-11-26
TargetHint Array throws Index Out Of Bounds Exception

pranaysharmamanulife:


hierynomus:
Double PR. Already merged with #588

Issue #603, opened 2021-01-26
File and Folder objects `getFilename()` returns null when enabledDFS is enabled on 0.10.1-SNAPSHOT


nddipiazza:
On 0.10.1-SNAPSHOT we are getting a null filename from the Folder and File objects

Please check out this branch: 

https://github.com/nddipiazza/smbj/tree/issue_603 / https://github.com/hierynomus/smbj/compare/master...nddipiazza:issue_603

And run it against any windows share:

```
./gradlew run  --args='-enableDfs -path "" -serverName 192.168.1.51 -share share -username administrator -domain nickis -password YOURPASSWORD'
```

returns a null for filename `result.getFileName()` when `-enableDFS` is enabled, and empty string when `-enableDFS` is not enabled. 

Instead it should have been set to the share smbpath: `\\192.168.1.51\share` regardless if `enableDfs` was true or false.

0.10.0 does *NOT* have this issue. 

This bug was introduced here: 

https://github.com/hierynomus/smbj/pull/350 / https://github.com/hierynomus/smbj/commit/f6997d3f0105ed6dcc3c06c4259187c5c358d0e5

hierynomus:
I've tried adding a reproduction testcase in #605, but at least there I do not get a `null` filename. Can you have a look to see what I'm missing?

nddipiazza:
Yeah I figured it out. Like I said this bug was introduced here https://github.com/hierynomus/smbj/commit/f6997d3f0105ed6dcc3c06c4259187c5c358d0e5

This used to give you a fileName that was the full UNC path:

https://github.com/hierynomus/smbj/blob/602346cdaa946f8d77e8d033e76d8596c5f8de50/src/main/java/com/hierynomus/smbj/share/DiskEntry.java#L57

How it has been changed to return the relative path:

https://github.com/hierynomus/smbj/blob/f6997d3f0105ed6dcc3c06c4259187c5c358d0e5/src/main/java/com/hierynomus/smbj/share/DiskEntry.java#L46

We should probably go back to UNC path 

```
    public String getFileName() {
        return name.toUncPath();
    }
```

Then if they want the Path they can use the `SmbPath.parse()` to get it. Or we can add a convenience method.

hierynomus:
Thanks for the fix!

PR #605, opened 2021-01-27
Tried adding reproduction testcase for #603

hierynomus:
@nddipiazza I've tried adding a reproduction testcase to the DFS integration test spec, but cannot seem to reproduce your issue, can you have a look?

### DIFF
--- a/src/it/groovy/com/hierynomus/smbj/DfsIntegrationSpec.groovy
+++ b/src/it/groovy/com/hierynomus/smbj/DfsIntegrationSpec.groovy
@@ -92,4 +92,21 @@ class DfsIntegrationSpec extends Specification {
     cleanup:
     share.close()
   }
+
+  def "should have filename for regular directory share when dfs is enabled GH#603"() {
+    given:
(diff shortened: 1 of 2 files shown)


nddipiazza:
HI I think it only happens when testing against windows server.

hierynomus:
Hmm... I don't have a windows server lying around unfortunately, do you have an ability to debug this issue and see where this is coming from?

hierynomus:
@pepijnve Do you have an ability to see where this is coming from?

nddipiazza:
We should sync up sometime. I can probably get you access to a windows VM that you can use for your integration tests.
I can debug and try to find the culprit when I get some time. I've got a bunch of high priority stuff on plate right now, next week probably earliest I can realistically take a look.

nddipiazza:
i took this branch and added the fix for it here: https://github.com/hierynomus/smbj/pull/610
updated the integration test as well

hierynomus:
Superseeded by #610

Issue #604, opened 2021-01-26
DFSPathResolver.sendDfsReferralRequest does not cache IPC$ share connections causing slowness


nddipiazza:
I notice that this is in a try-with-resources block. 

https://github.com/hierynomus/smbj/blob/accae9ea1a934e2755bff98bd26f4c0c54d9943e/src/main/java/com/hierynomus/smbj/paths/DFSPathResolver.java#L424

This makes it so that the DFS share resolution immediately removes this share from the `treeConnectTable` after getting the referral. 

The result of this is when DFS referrals are honored, smbj is many times slower because it constantly checks that over and over. 

Can we cache this and release share after a given time?

hierynomus:
Your PR has been merged!

PR #609, opened 2021-02-03
address issue #604 - stop closing the dfs share connection immediately.

nddipiazza:
Stop closing the `IPC$` share every time you connect to it to get a referral. 

In my use case where I am crawling a smb2 share, this resulted in 1000's of connections to `IPC$` every hour. 

Let these connections get cleaned up from the `logoff` method.

I tested this against a fairly massive distributed share and experienced no connection bloat.

perhaps it would be safer to keep these in a dfs share cache that expires if not used after `n` time and has a max cap of say 100 of these connections... but in my fairly massive use case it does not matter.


hierynomus:
Seems like a simple enough fix, though I wonder whether there would occur any problems from keeping the share open. I woould need to test it further.

PR #610, opened 2021-02-03
address issue #603

nddipiazza:
* address #603 where getFileName was returning path instead of unc path.
* deprecate getFileName - it's confusing. Instead offer a getPath and getUncPath method to DiskEntry.


hierynomus:
Great find!

hierynomus (approved):

PR #616, opened 2021-03-08
Issue-614: Add option to override asymmetric share

pranaysharmamanulife:
Issue-614: Add option to override asymmetric share

### DIFF
--- a/src/main/java/com/hierynomus/smbj/SmbConfig.java
+++ b/src/main/java/com/hierynomus/smbj/SmbConfig.java
@@ -77,8 +77,8 @@ public final class SmbConfig {
     private GSSContextConfig clientGSSContextConfig;
     private boolean encryptData;
     private String workStationName;
-
     private int soTimeout;
+    private boolean allowAsymmetricShares;
 
     public static SmbConfig createDefaultConfig() {
         return builder().build();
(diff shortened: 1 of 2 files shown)


hierynomus:
Adding support for _asymmetric` shares is a bit more involved than the implementation in this PR:

> If Connection.Dialect is "3.0.2" or "3.1.1" and the Capabilities field in the response includes the SMB2_SHARE_CAP_ASYMMETRIC bit, the client MUST verify whether both of the following conditions are true:
> - Connection.SessionTable contains only one entry.
> - Session.TreeConnectTable contains only one entry.
> If either of the preceding conditions is FALSE, the client MUST perform the following:
> - Disconnect the tree connection as specified in section 3.2.4.22.
> - Establish a new transport connection by providing the ServerName and TransportIdentifier used in the previous connection, as specified in section 3.2.4.2.1.
> - Send an SMB2 NEGOTIATE request on the new connection, as specified in section 3.2.4.2.2.2. The client also provides a newly generated Guid to be used as ClientGuid.
> - If the SMB2 NEGOTIATE request is successful, the client MUST create a new session on the new connection by sending an SMB2 SESSION_SETUP request, as specified in section 3.2.4.2.3. The client provides the UserCredentials used in the previous connection.
> - If the SMB2 SESSION_SETUP request is successful, the client MUST send an SMB2 TREE_CONNECT request, as specified in section 3.2.4.2.4. The client provides the ShareName used in the previous connection.
> - If the SMB2 TREE_CONNECT request is successful, the client SHOULD invoke the event as specified in [MS-SWN] section 3.2.4.1 by providing Connection.ServerName as the Netname parameter and TreeConnect.ShareName as the ShareName parameter, and by setting the IsShareNameNotificationRequired parameter to TRUE.

As such, I cannot merge it in its current state.

PR #625, opened 2021-03-25
Ensure session flags are set once session is initialised to prevent NPE

asmith-r7:
Established on session context is never called, meaning the session flags are never initialised on the session context which can lead to NPE when calling isGuest() or isAnonymous() on a session instance.

Based on code prior to 10.0.1-SNAPSHOT, I've added this right at the end of the session initialisation but happy to move elsewhere if deemed appropriate.


hierynomus (approved):

hierynomus:
Sorry for that oversight, added the `established` method and never calling it... 🤦

PR #631, opened 2021-05-11
Fix not derive signingKey in Smb3.x if not support/enable encryption

yin19941005:
Fix not derive signingKey in Smb3.x if not support/enable encryption

### DIFF
--- a/src/test/groovy/com/hierynomus/smbj/connection/SMBSessionBuilderSpec.groovy
+++ b/src/test/groovy/com/hierynomus/smbj/connection/SMBSessionBuilderSpec.groovy
@@ -15,10 +15,13 @@
  */
 package com.hierynomus.smbj.connection
 
+import com.hierynomus.mssmb2.SMB2Dialect
 import com.hierynomus.mssmb2.SMB3EncryptionCipher
+import com.hierynomus.mssmb2.messages.SMB2SessionSetup
 import com.hierynomus.protocol.commons.ByteArrayUtils
 import com.hierynomus.security.bc.BCSecurityProvider
 import com.hierynomus.smbj.SmbConfig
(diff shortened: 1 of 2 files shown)


yin19941005:
Here you go, I extracted the part as "deriveKeys" function for easier to implement the test.

src/test/groovy/com/hierynomus/smbj/connection/SMBSessionBuilderSpec.groovy
```diff
@@ -68,4 +71,48 @@ class SMBSessionBuilderSpec extends Specification {
...
+    def context = new SessionContext()
+    context.setSessionKey(new SecretKeySpec(sessionKey, HMAC_SHA256_ALGORITHM))
+
+    when:
+    def sessionBuilder = new SMBSessionBuilder(connection, config, null)
+    sessionBuilder.deriveKeys(response, SMB2Dialect.SMB_3_0, context)
+
+    then:
+    context.getSigningKey() != null
+    smb30xExpectedSigningKey == context.getSigningKey().encoded
```
hierynomus, line 94: Maybe add a check that the other keys are not derived in case of no encryption

hierynomus (changes requested):

yin19941005:
I guess that's what you mean: checking other keys in deriveKeys function. The application key is not checked as it don't have a get function yet (not ready to use, I guess).

src/test/groovy/com/hierynomus/smbj/connection/SMBSessionBuilderSpec.groovy
yin19941005: Is there any thing needed to add on this PR? I think the new commit already check other keys as well.

hierynomus (approved):

Issue #578, opened 2020-09-24
Thread getting stuck for more than 10 minutes during login


SubhasishDey:
Hi,

  We are finding that the SMBJ client stucks during connect for a period of 10 mins or more even if the socket timeouts are set in the SMBConfig.

  We see the below thread dump, any hint from anyone will be deeply appreciated.

Id=488 in TIMED_WAITING
	at sun.misc.Unsafe.park(Native Method)
	- waiting on <0x08e6d37d> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
	at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:215)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2163)
	at com.hierynomus.protocol.commons.concurrent.Promise.tryRetrieve(Promise.java:170)
	at com.hierynomus.protocol.commons.concurrent.Promise.retrieve(Promise.java:134)
	at com.hierynomus.protocol.commons.concurrent.PromiseBackedFuture.get(PromiseBackedFuture.java:58)
	at com.hierynomus.protocol.commons.concurrent.CancellableFuture.get(CancellableFuture.java:84)
	at com.hierynomus.protocol.commons.concurrent.Futures.get(Futures.java:38)
	at com.hierynomus.smbj.connection.Connection.multiProtocolNegotiate(Connection.java:340)
	at com.hierynomus.smbj.connection.Connection.negotiateDialect(Connection.java:310)
	at com.hierynomus.smbj.connection.Connection.connect(Connection.java:120)
	at com.hierynomus.smbj.SMBClient.getEstablishedOrConnect(SMBClient.java:103)
	- locked <0x04f57371> (a com.hierynomus.smbj.SMBClient)

ecki:
Can you show the other threads as well?

Without looking too deep into it, my experience in this area are broken DFS scenarios with dangling Server references. Did you try to target a server directly and turn DFS of?

SubhasishDey:
Hi ecki,

     Thanks for your reply, I have not set the dfs explicitly so I think it is false by default.
      But I will explicitly set it false and retry and look for any occurrence as this issue doesnot happen always.

Thanks and Regards,
Subhasish

hierynomus:
@SubhasishDey Any update on this?

SubhasishDey:
Hi @hierynomus / @ecki ,

[...]

PR #640, opened 2021-05-31
Only add DFSPathResolver if both client and server support DFS

hierynomus:


hierynomus:
This is an additional fix for issue #578

PR #650, opened 2021-07-02
Reducing logging for smb3

ndimitry:
For each smb3 packet there's an info log message which produces a tremendous amount of output.
I would suggest to reduce log level to debug (or trace; similarly as in one of the other packet reciever classes).


hierynomus (approved):
Thanks! Approved

ndimitry:
Thanks for the fast merging : )
I think in our tests we will use smb3 a lot, so in case something appears I will let you know! (But I hope all goes smoothly/)

ndimitry:
@hierynomus, from my side the same request: if we could have 0.11.2 that will be really great! : )

hierynomus:
0.11.3 is here (it skipped one due to an error in the release)

ndimitry:
Wow. that was fast! Thanks a lot!!

Issue #677, opened 2021-10-07
Hanging thread after `SocketTimeoutException` in `DirectTcpPacketReader`


heyalistair:
I'm getting an unrecoverable and hanging thread when there is a read timeout.

```
2021-09-24T16:16:58.622428896+01:00 stdout F  [30m2021-09-24 15:16:58,621 [0;39m  [34mINFO  [0;39m  [ [34mPacket Reader for ***************** [0;39m]  [33mcom.hierynomus.smbj.transport.PacketReader:53 [0;39m: PacketReader error, got exception.
2021-09-24T16:16:58.622428896+01:00 stdout F com.hierynomus.protocol.transport.TransportException: java.net.SocketTimeoutException: Read timed out
2021-09-24T16:16:58.622428896+01:00 stdout F 	at com.hierynomus.smbj.transport.tcp.direct.DirectTcpPacketReader.doRead(DirectTcpPacketReader.java:53)
2021-09-24T16:16:58.622428896+01:00 stdout F 	at com.hierynomus.smbj.transport.PacketReader.readPacket(PacketReader.java:70)
2021-09-24T16:16:58.622428896+01:00 stdout F 	at com.hierynomus.smbj.transport.PacketReader.run(PacketReader.java:48)
2021-09-24T16:16:58.622428896+01:00 stdout F 	at java.base/java.lang.Thread.run(Unknown Source)
2021-09-24T16:16:58.622428896+01:00 stdout F Caused by: java.net.SocketTimeoutException: Read timed out
2021-09-24T16:16:58.622428896+01:00 stdout F 	at java.base/sun.nio.ch.NioSocketImpl.timedRead(Unknown Source)
2021-09-24T16:16:58.622428896+01:00 stdout F 	at java.base/sun.nio.ch.NioSocketImpl.implRead(Unknown Source)
2021-09-24T16:16:58.622428896+01:00 stdout F 	at java.base/sun.nio.ch.NioSocketImpl.read(Unknown Source)
2021-09-24T16:16:58.622428896+01:00 stdout F 	at java.base/sun.nio.ch.NioSocketImpl$1.read(Unknown Source)
2021-09-24T16:16:58.622428896+01:00 stdout F 	at java.base/java.net.Socket$SocketInputStream.read(Unknown Source)
[...]

PR #701, opened 2022-03-14
https://github.com/hierynomus/smbj/issues/677

JozefDropco:
https://github.com/hierynomus/smbj/issues/677

this is a fix for above issue. We have faced the same with  our application. In case user making use of 

`try(SmbClient client: newClient()){  ... do usual work with connection and DFS share ...   }`

in our application we are not closing Connection ourselves but we were relying on SmbClient.close() method. Unexpectedly it didnt release all connection and PacketReader threads due the "leases" if the connection was reused several times it has decrement only once. At the end when we read the JavaDoc we found our that TransportLayer.disconnect should be called anytime.
`/* 
     * Close the Connection. If {@code force} is set to true, it forgoes the
     * {@link Session#close()} operation on the open sessions, and it just calls the
     * {@link TransportLayer#disconnect()}.`

Previous implementation was wrong as it returned without calling the transpartLayer.disconnect method

### DIFF
--- a/src/main/java/com/hierynomus/smbj/connection/Connection.java
+++ b/src/main/java/com/hierynomus/smbj/connection/Connection.java
@@ -168,10 +168,10 @@ public class Connection extends Pooled<Connection> implements Closeable, PacketR
      * @throws IOException If any error occurred during close-ing.
      */
     public void close(boolean force) throws IOException {
-        if (!force && !release()) {
-            return;
-        }
         try {
+            if (!force && !release()) {
+                return;
+            }
             if (!force) {
                 for (Session session : sessionTable.activeSessions()) {
                     try {


hierynomus:
Actually this change undoes the entire pooling. It will now always disconnect the transport layer, even when there are still other uses of the same connection.

hierynomus:
See also the failing test.

JozefDropco:
HI @hierynomus ,
yeah I got to know about this issue once I created Pull request, what do you propose?
There would be option to call from SmbClient.close method connection.close(true) but that would mean that the sessions are not closed gracefully but instead killed.

I tried to modify the code as little as possible to keep the contract same as it was but also ensure that calling just SmbClient.close will close all opened Threads

JozefDropco:
One more comment this test is not complete because its checking only if con2 isConnected().

`def "should not disconnect if more than one connection to the same host is open"() {
    given:
    def client = new SMBClient(config)
    def con1 = client.connect("hostA")
    def con2 = client.connect("hostA")

    when:
    con1.close()

    then:
    con1 == con2
    con2.isConnected()
  }`

 con1 should say its closed but will be reporting its still opened and developer can continue using con1. better would be create a proxy to real connection and in proxy keep the state.

hierynomus:
That might indeed simplify the reasoning about connections... I like the idea of handing out `ConnectionLease`s or something of that name, instead of the plain `Connection object.

JozefDropco:
Ok, lets say smbClient.getConnection will return ConnectionLease or LeasedConnection. Should still the Connection object count number of leases? Should ConnectionLease expose real connection by some getter? Should Connection be closeable from code or only by SmbClient.close? As far as I can see it will have quite and impact on code and also for other developers who had build their solution around bare Connection without contributing to SMBJ. I am not sure if I am the right one to do such rework of the code

Issue #718, opened 2022-06-16
SMBJ not working when Azure Files are used


JozefDropco:
As there is a code which does validation server details in SMBProtocolNegotiator for given "share" there must be always only one server which corresponds to the share name. This is not true for Azure Files where LRS (locally redundant storage) is selected. For consecutive requests for connection there might be different server responding (different GUID). We have faced it now. Connection as Poolable so it means whenever a connection is really closed SmbClient should unregister itself the server. Leaving it completely outside for user of this library is not easy to maintain the code itself. When connection to server is really closed a cleanup activity should be performed also on serverList.

JozefDropco:
I have read https://github.com/hierynomus/smbj/issues/644 but this seems to me as another problem where the connection is still opened and you try to do operations on open connection. In my case I am closing connection after "batch" of operations as long term connections are eating resources and should be avoided

PR #719, opened 2022-06-16
SmbClient cleanup when connection is getting closed serverList not cleared

JozefDropco:
Cleanup of SMBClient when connection is unregistered
https://github.com/hierynomus/smbj/issues/718


hierynomus (approved):

dheid:
@hierynomus A release would be so nice. We experience lots of trouble in our production systems, because of this issue.

Issue #730, opened 2022-08-27
Cannot connect to SMB shares on macOS machines


tdtran:
Authentication always fails when we try to connect to SMB shares on macOS machines. This is a regression. SMBJ used to work fine in 0.10.x series, but stopped working with 0.11.x.  I've narrowed down the cause of the regression to these two lines change in this one specific commit:

https://github.com/hierynomus/smbj/commit/0b6a5236b9c78c35151f1f3815931d68ec4bff86#diff-9ffb690b1078521e46395e84df6775742042ec2abf8e2f25e050e9033e9f8e65L124-R128

I am not knowledgeable enough regarding SMB protocol to tell why it does not work. Noneless reverting this change fixes the problem. Furthermore the new code contradicts the code comment. Someone changed the code but left the old comment.

I will submit a pull request.

hierynomus:
This is now fixed on the latest master.

dkocher:
> This is now fixed on the latest master.

Can you reference the changeset? It would also help to follow up if the _Milestone_ of closed issues is set to a release.

PR #731, opened 2022-08-27
Fix: unable to login macOS SMB shares

tdtran:
This is to fix issue #730 (Cannot connect to SMB shares on macOS machines)

Users who can't wait for the merge can use this jitpack.io build:

```
implementation 'com.github.tdtran:smbj:82b605041f0f81d6f38ec4ab659aa470a5698fa8'
```

I had to add another commit to tell jitpack to build this library with Java 11. It can't be built with the default Java 8.


hierynomus:
This is now fixed in the correct way on master. Closing this PR

PR #745, opened 2023-01-08
Change `mkdirs` to static

hma13:


ecki:
The idea is good, but it was already released. Do you want to break the API - in that case maybe a release notes is a good ideal.

also should the class get a private constrictor if all util methods will be static?

btw the recursive implementation has no provision for hitting the share as the parent, should it check for getParent() returning this?

hierynomus:
Indeed, oversight on my part, these should've been static from the get-go.

Maybe introduce a new static method that gets called from the instance methods, and mark those deprecated. That would maintain API compatibility and allow you to call the static variants directly.

ecki:
Maybe a new static until class with proper impl and Accessibility and delegate and deprecate the old class?

hierynomus:
Also makes sense to do! Let's go for that

PR #753, opened 2023-03-18
Fix Issue #745

hannosgit:
Attempts to fix issue #745 "Change mkdirs to static" with the implementation proposed in the comments of the PR.


hannosgit:
Should we also create the `copy` in the new util class?

neon-dev:
I was a bit hasty and created #757. I'd suggest implementing the fileIsPath boolean here.
Also, why not copy all the code over from `SmbFiles` and deprecate the old class entirely. This would also simplify deleting it later.

PR #754, opened 2023-03-19
Added List files

rokkakasu:
Added List files method to list all the files under the directory recursively using single session.

### DIFF
--- a/src/main/java/com/hierynomus/smbj/share/DiskShare.java
+++ b/src/main/java/com/hierynomus/smbj/share/DiskShare.java
@@ -274,6 +274,7 @@ public class DiskShare extends Share {
 							.filter(entry -> !entry.getFileName().equals("."))
 							.filter(entry -> !entry.getFileName().equals(".."))
 							.filter(entry -> !SmbFileAttributeUtils.isDirectory(entry.getFileAttributes()))
+							.map(info -> mapFullFilePath(path, info))
 							.collect(Collectors.toList()));
 				}
 			}
@@ -284,6 +285,11 @@ public class DiskShare extends Share {
         }
         return filesInformation;
     }
+
+	public FileIdBothDirectoryInformation mapFullFilePath(String path, FileIdBothDirectoryInformation info) {
+		info.setFullPath(path + "/" + info.getFileName());
+		return info;
+	}
     /**
      * Equivalent to calling {@link #list(String, Class, String, EnumSet<AccessMask>) list(path, informationClass, null, null)}.
      *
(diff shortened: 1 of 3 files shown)


hierynomus:
Hi @rokkakasu. Thank you for your effort. However, in this PR's current state, I won't merge it. Things like these do not belong in the core classes of the library. The core classes expose the base functionality. If I would merge this it would hide a huge performance penalty. These kind of things belong in classes built on top of the library. This could be added to for instance `SmbFiles`.

Also the code quality is well below standards of this repository. There are a lot of indentation errors.

rokkakasu:
Hi @hierynomus  Kindly reuse the part of the code I have provided so that it will be useful for many library users.
in present implementation to list files in each subfolders a separate NTLM auth is needed,
with this methods that can be avoided.
Kindly use the Code in any form.

Thanks,
@rokkakasu

laeubi:
@rokkakasu I think using a `Stream` would have many benefits, and is used in the NIO API as well, this will allow to iterate but still not eager fetch everything.

src/main/java/com/hierynomus/smbj/utils/SmbFileAttributeUtils.java
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hierynomus.smbj.utils;
+
+/**
+ * This utility class provides utilities for file attributes.
+ *
+ */
+public class SmbFileAttributeUtils {
```
laeubi, line 22: Could you maybe submit this as a separate PR to `SmbFiles` this is really useful and I missing such a good way to check if its a file or directory!

rokkakasu:
Sure.

rokkakasu:
Hi @hierynomus/smbj ***@***.***>

May i move this method from diskshare to Smbfiles class .

Will you approve my PR.

This utility can through all sub folders without creating new sessions for
each sub directories.

Thanks
@rokkakasu.

hierynomus:
Let me re-iterate. In the current implementation this will not be approved nor merged into the master branch. Submit a good PR, containing code, tests and proper formatting. Also not duplicating any constants already defined in the library, then I'll review and think about merging.

hierynomus:
As said initiallly, this code does not belong in the core classes of the library, but instead in a utilities class, such as `SmbFiles`.

src/main/java/com/hierynomus/smbj/share/DiskShare.java
```diff
@@ -239,7 +244,52 @@ public List<FileIdBothDirectoryInformation> list(String path) throws SMBApiExcep
...
+				FileIdBothDirectoryInformation directoryInformation = directoryIterator.next();
+				if (".".equalsIgnoreCase(directoryInformation.getFileName())
+						|| "..".equalsIgnoreCase(directoryInformation.getFileName())) {
+					continue;
+				}
+				if (SmbFileAttributeUtils.isDirectory(directoryInformation.getFileAttributes())) {
+					listAllFiles(path + "/" + directoryInformation.getFileName(), informationClass, searchPattern,
+							filesInformation);
+				} else {
+					filesInformation.addAll(directory.list(informationClass, searchPattern).stream()
```
MOubari0, line 272: Listing all elements in the directory here and then continuing to iterate over its elements on line 262 leads to duplicated values in the list. This else block should only add the current element to avoid duplication.

PR #772, opened 2023-07-05
Implement message integrity (ntlmssp_mic and mechListMic)

hierynomus:


hierynomus:
@Zwoca @jandornseifer Could you test this out? I think I solved the last missing piece in the authentication flow. Just need to double check I did not accidentally broke your environments after spending so much time debugging it ;)

JanDornseifer:
@hierynomus I can confirm that with 0.12.1 snapshot we can successfully connect, write and read files from customer's NAS server .

ZwoCa:
@hierynomus I can also confirm it's working fine, I checked it both with NtlmConfig#withIntegrity set to false/true. 👍

ryanfetz:
I just ran into this issue.  When is the plan to release 0.12.1 to maven central?

PR #783, opened 2023-08-11
Unregister cached server for same hostname

dheid:
Hi @hierynomus !

Our infrastructure team sometimes switches the machines behind the same hostnames. In this case, smbj throws an exception, because it keeps the servers in an in-memory list were they stay until an application restart. That forces us to restart the application.

I made a change so that no exception will be thrown but instead the cached server will be unregistered and newly registered if it changed on the same hostname. It would save us much restarting effort, if you could merge this patch. Thanks in advance!!!

Best wishes

Daniel

### DIFF
--- a/src/main/java/com/hierynomus/smbj/connection/SMBProtocolNegotiator.java
+++ b/src/main/java/com/hierynomus/smbj/connection/SMBProtocolNegotiator.java
@@ -229,7 +229,9 @@ class SMBProtocolNegotiator {
         } else if (temp.validate(cachedServer)) {
             negotiationContext.server = cachedServer;
         } else {
-            throw new TransportException(String.format("Different server found for same hostname '%s', disconnecting...", temp.getServerName()));
+            connection.serverList.unregister(temp.getServerName());
+            connection.serverList.registerServer(temp);
+            negotiationContext.server = temp;
         }
     }
 


hierynomus:
@dheid Thank you for the PR, however in its current form it is contrary to what the specification says:

```
3.2.5.2 Receiving an SMB2 NEGOTIATE Response
[...]
If the client implements the SMB 3.x dialect family, the client MUST look up the server entry in ServerList where Server.ServerName matches the Connection.ServerName. If an entry is found,
the client MUST set Connection.Server to the server entry found.
[...]
If the client implements the SMB 3.x dialect family and Connection.Server is not NULL, the client MUST disconnect the connection if any of the following conditions is satisfied:
 Connection.Server.ServerGUID does not match ServerGUID in the response.
 Connection.Server.DialectRevision does not match DialectRevision in the response.
 Connection.Server.SecurityMode does not match SecurityMode in the response.
 Connection.Server.Capabilities does not match Capabilities in the response.
```

What I think should happen is (this is not specified anywhere):
- When the last connection to the server disconnects, the server should be unregistered

Alternatively: Introduce a flag in the SmbConfig which disables the ServerList validation, and whose default setting is to be compatible with the spec.

dheid:
That's a good idea! I will do that as soon as I come back from my vacation

JozefDropco:
Server list cleanup Isnt it same as #719 which I have already provider a fix for?

hierynomus:
Yes @JozefDropco, it might well also solve the case of @dheid. I've just polished PR #719 now, and fixed the failing testcase there.

hierynomus:
@dheid Please take a look at the merged master. This incorporates #719, which probably should also solve your case.
If so, please let me know.

dheid:
@hierynomus Yes, this will solve our problem. We can close the PR.

dheid:
@hierynomus Will there be a new release soon? Asking for a friend ;-)

PR #795, opened 2023-09-05
Use a buffered inputstream to read from the socket

laeubi:
Currently the read from the socket is not buffered what can slow down reading bytes by byte.

This wraps the input stream in the PacketReader in a BufferedInputStream to read larger chuncks of data to improve performance.


laeubi:
@hierynomus I'm not sure if there are any performance tests but while profiling my application I noticed that most time is spend on reading from the socket so I would assume this can increase performance here.

hierynomus:
Thanks, I've fixed the styling issue detected, and approved and merged the PR.

PR #815, opened 2024-02-12
Enforce security by default

DHerls:
smbj by default does not require message signing and does not encrypt data. This could lead to program authors inadvertently writing insecure software.

The library should have secure defaults and allow users to reduce their security if they need.


hierynomus:
With these settings you'll also need to remove the supported dialects for SMB2. The goal of the default config is to make it more compatible. I would welcome a second "default setup" which is more secure.

So adding a `createSecureConfig()`

PR #823, opened 2024-05-02
Issue 817

nashkevichv:
Do not sign data packets (SMB2WriteRequest) if sining is not required (SmbConfig.signingRequired = false).


ecki:
This seems to be the default for windows client as well (sign if one side requires it) but there is also a “always” registry option. The builder should allow that as well (alwaysSign(true)) which would be the current behavior for 2.0+?

Issue #817, opened 2024-02-28
SMBJ Performance


proxymus89:
I noticed that using Windows file explorer I have speeds up to 4 times faster than smbj. why is smbj so slow? I tried to change all the possible parameters of smbj but to no avail. Is there something I'm missing or is it actually a limitation of the library?

hierynomus:
It highly depends on your method of testing... What types of streams are you using? Are you using the same dialect and encryption standards?

nashkevichv:
Thank you for the hint.
Below is output for the SmbFiles.copy(sourceFile, share, targetPath, true) try 
 
### smbJ with Smbfiles.copy
time [763] ms. speed [48472320] B/s size [36984380]
time [719] ms. speed [51438636] B/s size [36984380]
time [708] ms. speed [52237824] B/s size [36984380]
time [678] ms. speed [54549236] B/s size [36984380]
time [724] ms. speed [51083400] B/s size [36984380]
time [696] ms. speed [53138476] B/s size [36984380]
time [757] ms. speed [48856512] B/s size [36984380]
time [771] ms. speed [47969364] B/s size [36984380]
time [741] ms. speed [49911444] B/s size [36984380]
time [714] ms. speed [51798852] B/s size [36984380]
### smbJ with Smbfiles.copy 

overall performance is approximately the same.

nashkevichv:
I started test with all versions of smbj. Release 0.10.0 was 10% faster  than 0.13.0.

On pc with 10Gb network i got the following results :

### smbJ
buffer [12288] time [319] ms. speed [115938496] B/s size [36984380]
buffer [12288] time [302] ms. speed [122464832] B/s size [36984380]
buffer [12288] time [300] ms. speed [123281264] B/s size [36984380]
buffer [12288] time [298] ms. speed [124108656] B/s size [36984380]
buffer [12288] time [297] ms. speed [124526528] B/s size [36984380]
buffer [12288] time [290] ms. speed [127532344] B/s size [36984380]
buffer [12288] time [284] ms. speed [130226688] B/s size [36984380]
buffer [12288] time [285] ms. speed [129769760] B/s size [36984380]
buffer [12288] time [283] ms. speed [130686848] B/s size [36984380]
buffer [12288] time [285] ms. speed [129769760] B/s size [36984380]
### end of smbJ
### smbJ with Smbfiles.copy
time [440] ms. speed [84055408] B/s size [36984380]
time [447] ms. speed [82739104] B/s size [36984380]
[...]

PR #824, opened 2024-05-03
Performance improvement for unsigned packets (#817 fix)

manish-pai:
The Session.send() will check if the server has set RequireSecuritySignature in the Negotiate Protocol Response. If server and client do not wish to sign packets, an unsigned packet will be sent. 

This should address https://github.com/hierynomus/smbj/issues/817

### DIFF
--- a/src/main/java/com/hierynomus/smbj/session/Session.java
+++ b/src/main/java/com/hierynomus/smbj/session/Session.java
@@ -298,26 +298,37 @@ public class Session implements AutoCloseable {
         }
 
         ConnectionContext connectionContext = connection.getConnectionContext();
+        SMB2Dialect dialect = connection.getNegotiatedProtocol().getDialect();
         /*
-         * Ref: https://learn.microsoft.com/en-us/windows-server/storage/file-server/smb-signing-overview
-         * In summary SMB is signed when:
-         *    Both the SMB client and server have RequireSecuritySignature set to 1.
-         *    The SMB client has RequireSecuritySignature set to 1 and the server has RequireSecuritySignature set to 0.
-         *    The SMB server has RequireSecuritySignature set to 1 and the client has RequireSecuritySignature set to 0.
-         * Signing isn't used when:
-         *    The SMB client and server have RequireSecuritySignature set to 0.
+         * Ref: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/973630a8-8aa1-4398-89a8-13cf830f194d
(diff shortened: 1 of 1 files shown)


tooptoop4 (approved):

hierynomus:
It's always nice that the guys at MS write a spec (MS-SMB2) and then a tech-note, which to some extend contradict eachother...

```
The client MUST sign the message if one of the following conditions is TRUE:
- If Connection.Dialect is equal to "2.0.2" or "2.1", the message being sent contains a nonzero
value in the SessionId field and the session identified by the SessionId has
Session.SigningRequired equal to TRUE.
- If Connection.Dialect belongs to 3.x dialect family, the message being sent contains a nonzero
value in the SessionId field and one of the following conditions is TRUE:
- The session identified by SessionId has Session.EncryptData equal to FALSE.
- The tree connection identified by the TreeId field has TreeConnect.EncryptData equal to
FALSE.

If Session.SigningRequired is FALSE, the client MAY<99> sign the request.
```

hierynomus:
That means that your implementation is too simple unfortunately. It doesn't take into account the SMB3 requirements which are now implicitly covered because we just "always sign if we can"...

Also it feels like "sane" behaviour to sign when we can, and make it explicit in the client to not sign. Implicitly offering up security over performance is rarely a good idea.

manish-pai:
Thanks for the feedback. It basically means for SMB 3.x, either you encrypt or sign. I've incorporated the changes.

tooptoop4:
📦!

vairavlavy:
Is this PR not complete?

vairavlavy:
Please complete this PR ASAP as I need the exact same feature

manish-pai:
I believe that If the server does not expect client packets to be signed, it would not validate the integrity even if we sign it. Based on this assumption, I've made these changes.

tooptoop4:
:shipit:

vairavlavy:
Please complete this PR ASAP as I need the exact same feature

tooptoop4:
💥

tooptoop4:
💨

gpsmit:
@hierynomus what is the state of this PR by @manish-pai ? 
we have a corporate environment where windows also doesn't sign the packets, we are getting nowhere near the speeds regular windows clients are getting because this library always signs regardless of our requests not to sign

gpsmit:
@hierynomus could you at least respond why or why you aren't satisfied yet after the last changes? Then @manish-pai could adjust it once more?

hierynomus:
I've merged a slightly different fix for this in #848

PR #835, opened 2024-06-25
Try all resolved IP addresses for hostname (Fixes #834)

### LINKED ISSUE #834: Connection Timeout : Connect to DNS with multiple IPs with active / standby mechanim
Hi Team,

I want to connect to FSx Multi AZ in AWS through aliases / DNS.
Try nslookup, this DNS return 2 IPs (active and standby), basically only active server can be accessed.

Beause DNS round-robin behaviour.
It will works if connect to active server / IP, else error "socket conection timeout". See picture below.

Through netcat, it can connect. While it come to standby IP it will try to connect and then go to active IP.
image

Application already tested 0.13.0 
image

There is any way to handle kind of this issue ?

Thanks

hierynomus:
Signed-off-by: Jeroen van Erp <jeroen@hierynomus.com>

### DIFF
--- a/src/main/java/com/hierynomus/protocol/commons/HostResolver.java
+++ b/src/main/java/com/hierynomus/protocol/commons/HostResolver.java
@@ -13,17 +13,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.hierynomus.smbj.testing;
+package com.hierynomus.protocol.commons;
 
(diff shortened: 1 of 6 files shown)


danielhasugian (approved):
ok

Nicomko9:
Can we revive this ? Make sure the build passes to integrate that fix

PR #850, opened 2024-12-11
Fix InputStreamByteChunkProvider#isAvailable

vorobii-vitalii:
Hello!

There is issue in this function - https://github.com/hierynomus/smbj/blob/b78c87285bcec7b18895e1ccb46d51b9f22f649e/src/main/java/com/hierynomus/smbj/io/InputStreamByteChunkProvider.java#L53.
According to docs - https://docs.oracle.com/javase/8/docs/api/java/io/BufferedInputStream.html. BufferedInputStream#available returns only estimate of the number of bytes that can be read.
And one of implementations of InputStream we rely on actually always returns zero before reading.
Please review my approach. I've tested it and it works as expected

### DIFF
--- a/src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
+++ b/src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright (C)2016 - SMBJ Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.hierynomus.smbj.io;
 
 import java.io.BufferedInputStream;
--- a/src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
+++ b/src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
@@ -1,0 +1,33 @@
+package com.hierynomus.smbj.io;
+
+import java.io.BufferedInputStream;
+import java.io.IOException;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class BufferedInputStreamReader {
+    private static final int EOF = -1;
+
+    private final AtomicBoolean isEnd = new AtomicBoolean(false);
+
+    private final BufferedInputStream inputStream;
+
+    public BufferedInputStreamReader(BufferedInputStream inputStream) {
+        this.inputStream = inputStream;
+    }
+
+    public int read(byte[] byteArray, int offset, int length) throws IOException {
+        int readResult = inputStream.read(byteArray, offset, length);
+        if (readResult == EOF) {
+            isEnd.set(true);
+        }
+        return readResult;
+    }
+
+    public boolean isAvailable() {
+        return !isEnd.get();
+    }
+
+    public void close() throws IOException {
+        inputStream.close();
+    }
+}
--- a/src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
+++ b/src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
@@ -17,12 +17,14 @@ package com.hierynomus.smbj.io;
 
 import java.io.BufferedInputStream;
 import java.io.IOException;
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import com.hierynomus.smbj.common.SMBRuntimeException;
 
 public class BufferedInputStreamReader {
     private static final int EOF = -1;
 
-    private final AtomicBoolean isEnd = new AtomicBoolean(false);
+    private final AtomicInteger cachedByte = new AtomicInteger(EOF);
 
     private final BufferedInputStream inputStream;
 
@@ -31,15 +33,27 @@ public class BufferedInputStreamReader {
     }
 
     public int read(byte[] byteArray, int offset, int length) throws IOException {
-        int readResult = inputStream.read(byteArray, offset, length);
-        if (readResult == EOF) {
-            isEnd.set(true);
+        if (cachedByte.get() != EOF) {
+            byteArray[offset] = (byte) cachedByte.getAndSet(EOF);
+            int readResult = inputStream.read(byteArray, offset + 1, length - 1);
+            return readResult == EOF ? 1 : (readResult + 1);
+        } else {
+            return inputStream.read(byteArray, offset, length);
         }
-        return readResult;
     }
 
     public boolean isAvailable() {
-        return !isEnd.get();
+        try {
+            if (inputStream.available() > 0 || cachedByte.get() != EOF) {
+                return true;
+            }
+            int readByte = inputStream.read();
+            cachedByte.set(readByte);
+            return readByte != EOF;
+        }
+        catch (IOException e) {
+            throw new SMBRuntimeException(e);
+        }
     }
 
     public void close() throws IOException {
--- a/src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
+++ b/src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
@@ -17,26 +17,28 @@ package com.hierynomus.smbj.io;
 
 import java.io.BufferedInputStream;
 import java.io.IOException;
-import java.util.concurrent.atomic.AtomicInteger;
 
 import com.hierynomus.smbj.common.SMBRuntimeException;
 
 public class BufferedInputStreamReader {
     private static final int EOF = -1;
 
-    private final AtomicInteger cachedByte = new AtomicInteger(EOF);
-
     private final BufferedInputStream inputStream;
+    private volatile int cachedByte = EOF;
 
     public BufferedInputStreamReader(BufferedInputStream inputStream) {
         this.inputStream = inputStream;
     }
 
     public int read(byte[] byteArray, int offset, int length) throws IOException {
-        if (cachedByte.get() != EOF) {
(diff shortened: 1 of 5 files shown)


vorobii-vitalii:
@hierynomus Any news??

laeubi:
@vorobii-vitalii As you noticed using the `isAvailable` for anything really is unreliable I wonder where/who it is used, maybe one better fix that place to not use `isAvailable` at all...

vorobii-vitalii:
Hi, @laeubi 
I agree that relying on BufferedInputStream#available is a bad idea, since it returns estimate.
However, if we want to read from some input stream in chunks and transfer chunks to Samba we still need to know whether underlying ByteChunkProvider can still provide some bytes. We shouldn't exit following loop until we are sure ByteChunkProvider won't provide any bytes.
```
public long write(ByteChunkProvider provider, ProgressListener progressListener) {
        int bytesWritten = 0;
        **while (provider.isAvailable()) {**
            logger.debug("Writing to {} from offset {}", this.entryName, provider.getOffset());
            SMB2WriteResponse wresp = share.write(fileId, provider);
            bytesWritten += wresp.getBytesWritten();
            if (progressListener != null)
                progressListener.onProgressChanged(wresp.getBytesWritten(), provider.getOffset());
        }
        return bytesWritten;
    }
```
Perhaps function name can be improved

vorobii-vitalii:
And I am not sure we can avoid this polling loop, we have to read from file/socket in chunks, because otherwise we might run into out of memory error

laeubi:
> I agree that relying on BufferedInputStream#available is a bad idea, since it returns estimate.

It does not, per javadoc it *might* return an estimate it is also legal to just return 0 and this is the case you are encountering here.

So this PR actually makes `isAviable` to something like "wasEOF read" what already is part of the API by returning `-1` ... 

> And I am not sure we can avoid this polling loop, we have to read from file/socket in chunks, because otherwise we might run into out of memory error

The **provider** need to know if there is more to *write* but this should not result in using available on an input stream! Instead it should work similar to how `Iterator` works, that is a call to `hasNext()` will actually perform a read of the next data and cache it somewhere or return false (in this case e.g. it has read EOF). then `next()` can use this cached data (what might be a small chunck) to write it / return it / whatever...

src/main/java/com/hierynomus/smbj/io/InputStreamByteChunkProvider.java
```diff
@@ -43,28 +39,24 @@ int prepareChunk(byte[] chunk, int bytesNeeded) throws IOException {
...
     }
 
     @Override
     public boolean isAvailable() {
-        try {
-            return super.isAvailable() || (is != null && is.available() > 0);
-        } catch (IOException e) {
-            throw new SMBRuntimeException(e);
-        }
+        return super.isAvailable() || (bufferedInputStreamReader != null && bufferedInputStreamReader.isAvailable());
```
laeubi, line 47: As we know for sure this is a Buffered Stream, one could do a `mark(1)` operation, then read one byte and then `reset` if `is.available()` is returning `0`.

That way at least one byte will be queried and it could be checked if EOF was reached.

vorobii-vitalii: Thanks for suggestion, I implemented very similar solution. Please have a look

vorobii-vitalii:
> > I agree that relying on BufferedInputStream#available is a bad idea, since it returns estimate.
> 
> It does not, per javadoc it _might_ return an estimate it is also legal to just return 0 and this is the case you are encountering here.
> 
> So this PR actually makes `isAviable` to something like "wasEOF read" what already is part of the API by returning `-1` ...
> 
> > And I am not sure we can avoid this polling loop, we have to read from file/socket in chunks, because otherwise we might run into out of memory error
> 
> The **provider** need to know if there is more to _write_ but this should not result in using available on an input stream! Instead it should work similar to how `Iterator` works, that is a call to `hasNext()` will actually perform a read of the next data and cache it somewhere or return false (in this case e.g. it has read EOF). then `next()` can use this cached data (what might be a small chunck) to write it / return it / whatever...

I agree that my solution changes semantics of this function, implemented different approach

src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
```diff
@@ -31,15 +33,27 @@ public BufferedInputStreamReader(BufferedInputStream inputStream) {
...
+        } else {
+            return inputStream.read(byteArray, offset, length);
         }
-        return readResult;
     }
 
     public boolean isAvailable() {
-        return !isEnd.get();
+        try {
+            if (inputStream.available() > 0 || cachedByte.get() != EOF) {
```
laeubi: I would first check for the cached byte, this is probably always faster, also its easier to read/debug/understand...
```suggestion
            if (cachedByte.get() != EOF) {
                 return true;
            }
            if (inputStream.available() > 0) {
```

src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
```diff
@@ -31,15 +33,27 @@ public BufferedInputStreamReader(BufferedInputStream inputStream) {
...
+                return true;
+            }
+            int readByte = inputStream.read();
+            cachedByte.set(readByte);
+            return readByte != EOF;
```
laeubi:
```suggestion
            return readByte > EOF;
```

src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
```diff
@@ -17,12 +17,14 @@
...
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import com.hierynomus.smbj.common.SMBRuntimeException;
 
 public class BufferedInputStreamReader {
     private static final int EOF = -1;
 
-    private final AtomicBoolean isEnd = new AtomicBoolean(false);
+    private final AtomicInteger cachedByte = new AtomicInteger(EOF);
```
laeubi:
```suggestion
    private volatile int cachedByte = EOF;
```

I don't think we can really make use of an atomic value here, a simple volatile variable should serve the same purpose, ready concurrently from different threads from a stream looks like waiting for ba things to happen, especially in combination with non atomic calls like aviable / read ...

src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
```diff
@@ -31,15 +33,27 @@ public BufferedInputStreamReader(BufferedInputStream inputStream) {
     }
 
     public int read(byte[] byteArray, int offset, int length) throws IOException {
-        int readResult = inputStream.read(byteArray, offset, length);
-        if (readResult == EOF) {
-            isEnd.set(true);
+        if (cachedByte.get() != EOF) {
+            byteArray[offset] = (byte) cachedByte.getAndSet(EOF);
+            int readResult = inputStream.read(byteArray, offset + 1, length - 1);
+            return readResult == EOF ? 1 : (readResult + 1);
```
laeubi:
```suggestion
           if (length == 1) {
              return 1;
           }
            int readResult = inputStream.read(byteArray, offset + 1, length - 1);
            return readResult > EOF ?  (readResult + 1):  1;
```

laeubi (commented):
I think this is a much better approach, just some style comments.

src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
vorobii-vitalii: Branch prediction would help I guess, but I swapped conditions anyway

src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
vorobii-vitalii: Changed

src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
vorobii-vitalii: Yeah, volatile is enough. Actually it appears this object is not used across multiple threads so in theory we don't even need volatile, but better safe than sorry.

src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
vorobii-vitalii: Changed

hierynomus:
Before I merge this change in. I'd like to see a testcase that proves the original implementation wrong.

laeubi:
@hierynomus there is already a testcase in this PR, @vorobii-vitalii does this testcase fails without your patch?

hierynomus:
I've seen the testcase indeed, I'm not convinced it fails without.

vorobii-vitalii:
It was test case for new component. Added new one 
With previous implementation it fails:
image
With new it passes:
image

vorobii-vitalii:
@hierynomus @laeubi Hey guys! Is PR still missing some changes required for merge?

src/main/java/com/hierynomus/smbj/io/BufferedInputStreamReader.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.hierynomus.smbj.common.SMBRuntimeException;
+
+public class BufferedInputStreamReader {
+    private static final int EOF = -1;
+
+    private final BufferedInputStream inputStream;
+    private volatile int cachedByte = EOF;
+
+    public BufferedInputStreamReader(BufferedInputStream inputStream) {
```
laeubi, line 29: Does it really requires a `BufferedInputStream` here? If not then a plain `InputStream` would make this clas more reusable

laeubi (approved):
It looks good to me but I'm not a committer on the project :-)

vorobii-vitalii:
Hello @hierynomus  Please let me know what changes are missing in this PR.

hierynomus:
Thanks for identifying and fixing this real bug — [available()](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is indeed unreliable and the existing code was broken for any [InputStream](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) that always returns 0 from it.

After reviewing the PR I think there's a simpler approach that avoids the BufferedInputStreamReader wrapper entirely and also fixes a subtle secondary issue with your implementation.

I've fixed this in 425449e733c55ccf7495ec1f030096d018159f45

PR #859, opened 2025-01-31
java.nio.FileSystem support

dgodfrey:
This mainly adds support for reading from SMB shares using a FileSystem implementation. It use the URI format from jcifs, to ease migrations.

It was developed on Windows, so hopefully is ok on Linux, and includes a few "fixes" to SmbStubServer for Windows and a change so it only rebuilds the docker image used for integration testing once, rather than every times it's used.


laeubi:
@dgodfrey I think it would help if you squash all commit into one here.

@hierynomus any chance to make progress here? I also using a private fork of this code since a while and like to see this in the main repo and contribute some additional fixes and enhancements to this.

dgodfrey:
I don't really like squash commits, as it doesn't tell the "story", although I suppose it won't make too much difference here as 95% of the files are new. I don't really know Github - is it possible to squash at merge time?

laeubi:
@dgodfrey yes you can do it in Github UI, while I agree that the history is often valuable for a PR it is often better to have a single commit (or maybe two) maybe done in a separate branch as it is easier to rebase if the repository code changes.

In any case @hierynomus has of course to tell what is the best here, it was just a thought as I see 30 commits are here for a single feature :-)

src/main/java/com/hierynomus/smbj/share/File.java
```diff
@@ -195,9 +195,10 @@ public void read(OutputStream destStream, ProgressListener progressListener) thr
      * @param fileOffset The offset, in bytes, into the file to which the data should be written
      * @return the actual number of bytes that was written to the file
      */
-    public long write(ByteBuffer buffer, long fileOffset) {
+    public int write(ByteBuffer buffer, long fileOffset) {
```
laeubi, line 198: I think this should not be truncated here, if the caller wants an int it is better cast there.

src/main/java/com/hierynomus/smbj/share/File.java
```diff
@@ -208,7 +209,7 @@ public long write(ByteBuffer buffer, long fileOffset) {
      * @param fileOffset The offset, in bytes, into the file from which the data should be read
      * @return the actual number of bytes that were read; or -1 if the end of the file was reached
      */
-    public long read(ByteBuffer buffer, long fileOffset) {
+    public int read(ByteBuffer buffer, long fileOffset) {
```
laeubi, line 212: Same here better use a long as long as possible.

dgodfrey: This was getting widened from an int to a long here, then any callers were needing to narrow it back to an int - this just removes that need. Pretty much the whole IO library uses int's for reading/writing bytes (I guess) because that's the maximum size of an array. But again, pretty easy on int or long.

src/main/java/com/hierynomus/smbj/share/File.java
dgodfrey: I think I changed them as the maximum size of a ByteBuffer is an int, so narrowed it here to match the read version and to do it once, rather than everybody who calls the method having to do it. But pretty easy either int or long.

dgodfrey:
Hi - @hierynomus do you have time to look at this - or would I be better putting it into a separate library?

dkocher:
> Hi - @hierynomus do you have time to look at this - or would I be better putting it into a separate library?

I am not a maintainer but I would prefer having this in a separate library keeping _smbj_ lean.

laeubi:
> > Hi - @hierynomus do you have time to look at this - or would I be better putting it into a separate library?
> 
> I am not a maintainer but I would prefer having this in a separate library keeping _smbj_ lean.

I don't see that this adds anything as it all is plain java API, having it directly in smbj (given its need for tight integration) makes sense as otherwise there is a need to maintain a stable API or expose internal accessors.

hierynomus (approved):

hierynomus:
Finally had some time to review and merge this. For the sake of speed, I've fixed the issues I found myself. Have a look through them.

laeubi:
@hierynomus many thanks!

laeubi:
@hierynomus I now tried to port some of my changes / improvements of the fs to what was actually merged, I started with two rather simple things before going on to more complex ones:

- https://github.com/hierynomus/smbj/pull/886 (mostly a bit house-keeping)
- https://github.com/hierynomus/smbj/pull/887 (performance improvement)

laeubi:
This one is based on the previous one and adds true `FileChannel` support (needed by some of the Files API functions):

- https://github.com/hierynomus/smbj/pull/888

laeubi:
This is now the most noteable enhancement, having `WatchService` support what prevents one from constant polling eventually and still using high-level java `Path` API:

- https://github.com/hierynomus/smbj/pull/889

PR #886, opened 2026-07-27
smbfs: add serialVersionUID and tighten InvalidShareException visibility

laeubi:
- InvalidShareException/ToBeImplementedException are missing serialVersionUID (Serializable via RuntimeException/UnsupportedOperationException).
- InvalidShareException's constructor is only ever used within com.hierynomus.smbfs; tighten to package-private.


laeubi:
The failure seems

```
DirectoryListingIntegrationTest > listsDirectories(String, List) > [2] dfs/public, [\public\folder, \public\test.txt] FAILED
    org.opentest4j.AssertionFailedError: expected: <[\public\folder, \public\test.txt]> but was: <[\public\test.txt, \public\folder]>
```

what look like an ordering problem and I can't really map this to my changes here...

PR #892, opened 2026-09-05
fix: Scope directory lease keys and cache to UNC path

courville:
Hi, I am the main maintainer of the Nova Video Player Android application. Nova uses several SMB implementations, one of which is smbj. While troubleshooting https://github.com/nova-video-player/aos-AVP/issues/1927, I tracked down a directory leasing issue in smbj. I added a test failing without the fix and passing with the proposed code change.

### Problem Description
`LeaseManager` is connection-scoped and shared across all shares connected on the same server IP.

Previously, `DiskShare` registered and looked up directory leases and cached listings using only the relative share path (`full.getPath()`), which evaluates to an empty string `""` for the root of any share.

Consequently:
1. **Cross-share cache pollution / directory collision:** Opening the root of one share (e.g. `\\host\Movies`) cached its listing in `LeaseManager` and `LeasedDirectoryCache` under `""`. Subsequent access to the root of a different share on the same server connection (e.g. `\\host\Paylasim`) hit the cache under `""` and immediately returned the first share's files without sending any network request.
2. **Invalid parent lease key:** When opening a share root (`path = ""`), `full.getParent()` returned the same path `""`. This mistakenly caused the root directory to query and set its own lease key as its `parentKey` in the `SMB2LeaseCreateContext.v2` request.

### Evidence / Reproduction Test
Added a unit test in `DiskShareListSpec.groovy` that connects to two shares (`"Movies"` and `"Paylasim"`) over the same connection session and opens their root directories:
- **Before this fix (unmodified master):** The test fails immediately because both shares generate and request the exact same `LeaseKey` for their roots:

```
  Condition not satisfied:

  requestedLeaseKeys[0] != requestedLeaseKeys[1]
  |                 |   |  |                 |
  |                 |   |  |                 LeaseKey{f0e944f479514ecaa3ab4e6558e1838b}
  |                 |   |  [LeaseKey{f0e944f479514ecaa3ab4e6558e1838b}, LeaseKey{f0e944f479514ecaa3ab4e6558e1838b}]
  |                 |   false
  |                 LeaseKey{f0e944f479514ecaa3ab4e6558e1838b}
```

- **After this fix:** The test passes cleanly: each share requests a distinct lease key scoped by share UNC path, and neither share sets an invalid self-referencing parent lease key.

### Remedy
1. **Scope to UNC path:** In `DiskShare.java`, `openDirectoryWithLease` and `list` now use `full.toUncPath()` (`leasePath`) for `LeaseManager.leaseKeyForPath()`, `LeaseManager.getByPath()`, `LeaseEntry`, and `LeasedDirectoryCache.serve()` / `.populate()`.
2. **Guard parent key lookup:** Only query and set `parentKey` when `!rel.isEmpty()`, ensuring a share's root has `parentKey = null`.
3. **Tests:** Updated `LeaseManagerSpec.groovy` and added multi-share root verification in `DiskShareListSpec.groovy`. Full test suite passes cleanly.


hierynomus:
Great find, thanks for the PR, I'll have a look

courville:
Thx. Am I required to change the PR title?

